PR #272, opened 2014-12-28
Fixing small 'contextual' typo

devopsmariocom:
Hope I'm not wrong, so either missing noun or just simply `context` should work.

Thanks for your work!


elmariofredo:
Ahhh got it, sorry I should not try to fix typos early in the morning. Thanks for your clarification!

getify:
no worries :)

PR #275, opened 2015-01-03
removes function's name of anonymous function

SotirisValogiannis:


getify:
errata submitted to o'reilly

PR #280, opened 2015-01-06
Update ch1.md

Praveen-Innocent:


getify:
The first chapter will not be called the same as the book title. And while those two topics listed will indeed be covered, it's unlikely they'll be in chapter 1. Please consider asking before submitting new content to find out what the plans are.

PR #299, opened 2015-01-12
"this & object prototypes": using new empty array to apply concat method

fedosejev:
Seems like it doesn't make a difference whether you use existing `curried` array or a new one to apply `concat` method, but it might save the confusion and make this example easier to understand. Please let me know if I am missing something.

### DIFF
--- a/this & object prototypes/ch2.md
+++ b/this & object prototypes/ch2.md
@@ -728,7 +728,7 @@ if (!Function.prototype.softBind) {
 						(typeof global !== "undefined" &&
 							this === global)
 					) ? obj : this,
-					curried.concat.apply( curried, arguments )
+					[].concat.apply( curried, arguments )
 				);
 			};
 		bound.prototype = Object.create( fn.prototype );


getify:
> Seems like it doesn't make a difference

Probably not a *big* difference, but in your version it might be allocating an `[ ]` empty array only to throw it away for GC.

But more importantly, I'm not sure why `[].concat...` is less confusing... the point of that line is to concatenate onto the `curried` array, so `curried.concat(..)` would be the thing we'd normally want to call.

Unfortunately, `arguments` isn't really an array so we can't just do `curried.concat(arguments)`. So we use a little `apply(..)` trick to spread out `arguments` into params to `concat`, which lets the concatenation happen.

I don't think using the `apply(..)` part obscures the fact that whatever we're doing is "concatenating onto 'curried' array", which is what `curried.concat...` is supposed to imply.

Am I missing some angle that made `curried.concat` more confusing to you than `[].concat`?

alexreardon:
Rather than using ``[].contact.apply`` and relying on the garbage collection, perhaps it is worth substituting ``Array.prototype.concat.apply`` to avoid junk array creation.

getify:
I guess I'm missing something because that just seems even more confusing. The point of the code is to concatenate `arguments` onto `curried`. So how is `curried.concat..(..)` not the most sensible of the various options?

fedosejev:
Kyle, thanks for explaining in more details.

Good point on GC.

> I don't think using the `apply(..)` trick obscures the fact that whatever we're doing is "concatenating onto 'curried' array", which is what `curried.concat...` is supposed to imply.

That's where I got confused. The `apply(..)` trick made immediate sense, but then `curried.concat...` got me thinking that there is something else going on that I don't understand. Now when you've explained that the sole purpose of `curried.concat..(..)` is to imply that we're "concatenating onto 'curried' array" - I can see why it's more verbose over `Array.prototype.concat(..)`.

I guess it would avoid the confusion (at least for me) by breaking that line into 2 steps: 1) use `Array.prototype.concat(..)` to get it working, and then 2) refactor it with `curried.concat..(..)` to get it more verbose.

Thanks for making it clear.

cobalamin:
The problem, in short, seems to be that the `curried.concat` part is merely grabbing a reference to the `concat` function that is actually on the `Array.prototype` and not specific to the `curried` array. I understand that this can be confusing.

To use `Function.prototype.apply`, we need to get a reference to some function. We then call "apply" on it with the first argument being what would usually come before the dot -- the affected object.

So if we're using the object `curried` itself to just grab a reference to the globally unique function `Array.prototype.concat`, and then `apply()` that function with `curried` as the first argument/*this*, I can easily see how the two `curried`s lead to confusion.

`Array.prototype.concat.apply(curried, arguments)` seems verbose, but in my opinion it's by far the better choice, as it avoids confusion and makes clear what we're really doing with `apply` here.

getify:
Sorry, not trying to be difficult, but I just don't see/understand the argument that in comparing these two, the former is better (less confusing) than the latter:

```js
Array.prototype.concat.apply( curried, arguments )
curried.concat.apply( curried, arguments )
```
**Note:** They both require `contact.apply( curried, arguments )`. The difference is `Array.prototype` vs `curried`.

Keep in mind the context... this code snippet appears early in the book, chapter 2, when all we're talking about so far is the `this` keyword. It's not until several chapters later that we explain the concept of `.prototype`, object prototypes, etc. I think introducing `Array.prototype` into the example adds extra detail/complication.

I appreciate both of your inputs here. I guess I just see it opposite.

cobalamin:
From the perspective of someone without knowledge of prototypes, true. The example stays somewhat confusing to me, but IMHO it's still more comprehensible than `[].concat...`. However, we're already doing prototype stuff like

>`bound.prototype = Object.create( fn.prototype );`

in the very same snippet.

Also, why are you using the `[].` syntax just a few lines above in the same snippet?

>`curried = [].slice.call( arguments, 1 ),`

In this example, you've probably chosen that syntax over `curried = curried.slice.call`, which would also seem tremendously confusing to me... Not an easy topic.

getify:
> we're already doing prototype stuff like `bound.prototype = Object.create( fn.prototype );`

You're right. In truth, the snippet itself is more complex than I'd like, and more complex than I spend the time to fully explain. You'll notice the surrounding text doesn't really get into explaining the details of how `softBind(..)` works... I just present it as an as-is utility that you can use or not if you want. It would take another page or two of explanatory text to fully flush out how and why it does what it does internally.

So, in *this* perspective, I guess some of our earlier philosophical discussions about simplicity vs complexity are a little ... stretched. In that PoV, it would probably not matter to the overall snippet if I use `curried.concat` or `Array.prototype.concat`, as they're both surrounded by a lot of other complexity that I'm also not explaining.

The only observation I can make along these lines is that this snippet wasn't just written only for the book. That code, as presented (except for some line formatting that was book-specific) is how I use `softBind(..)` in my own code. So the choices I made about `curried.concat` for semantic hinting were more for *that* code usage and less about what was specifically presented in this book.

> Also, why are you using the []. syntax just a few lines above in the same snippet? 
> `curried = [].slice.call( arguments, 1 )`
> In this example, you've probably chosen that syntax over curried = curried.slice.call, which would also seem tremendously confusing to me... Not an easy topic.

This is a good point and a good question.

But here, it's subtly but importantly different than the original topic here. For `[].slice.call( arguments, 1 )`, you'll notice that `arguments` could not replace `[]` (to make any sort of semantic hint as I did with `curried.concat`) because `arguments` isn't a real array. The whole reason we're doing the `.call(..)` indirection here is to work-around the crappiness that `arguments` isn't a real array. If it were, we wouldn't do `arguments.slice.call( arguments, 1)`, we'd just do `arguments.slice( 1 )`.

So I guess my answer is, I use `[].slice.call` because I can't use `arguments.slice` and because `Array.prototype.slice.call` is, again, I think unnecessarily longer.

-------------

Another observation: this book has already been published 6+ months ago, and was in online form for nearly another 6 months before that. So the text here is already a year old. And it's the first edition of the book.

Any change we're suggesting making here wouldn't necessarily be appropriate as an errata for the book (unless it was a bug), but would instead be appropriate for a second edition tweak.

In that sense, I could entertain the idea of whether the whole snippet needs its own detailed explanation. I probably wouldn't put that content directly into the chapter, as it breaks up the discussion of `this` unnecessarily. But it could have been in an appendix that I noted and linked to.

I don't know if I'm going to do this, but it's at least worth considering. So I'm labeling this issue as "second edition" and will revisit it at a later time when it's time to revise the book.

Thanks for all the input and critical thinking! :)

PR #310, opened 2015-01-20
"Finally" section contained a syntax error

benjamingr:
Removed a semicolon. I wasn't sure if chaining was the intended behaviour or a "p" was missing from the second call so I just picked the one that made the most sense. Cheers.


getify:
Manually fixed this in my latest updates. Thanks for the catch!

benjamingr:
@getify thanks for the book! It's turning out quite nicely!

getify:
thanks, appreciate your comments and review! :)

PR #313, opened 2015-01-21
Use genitive for "your being able to write"

rsperberg:
Rule 6.21 in Chicago Manual of Style (14th ed.) specifies the genitive case (which is "analogous to possessives and formed like them") in such constructions.


getify:
Thanks!

PR #327, opened 2015-02-11
types+grammar: fix typos in update foreword

cvan:
to be consistent with

> "gotchas,"

although I think chapters should be in quotation marks only, not italicised - but it depends on which style guide you're following.


getify:
Good catch, this title had not had the style updated to exclude the `"` quotes around the book titles... only leaving italics. Thanks! :)

PR #328, opened 2015-02-12
remove duplicated article

maxhoffmann:
I wasn’t sure which one you’d like to use. Went with `the` in this pull request.


benjamingr:
Well, the semantic issue is whether or not that object has a single or multiple JSON stringifications. I think either are ok.

getify:
The book had already been changed to include "a", so I fixed this merge to pick that one. :) Thanks for the catch!

PR #347, opened 2015-02-23
Update ch1.md

Cmdv:
Incorrect results needed amending


getify:
I'll keep this around for possible update in second edition. Thanks.

PR #351, opened 2015-03-01
Talk about template strings

mallowigi:
Some rephrasing might be necessary, I am not an eloquent writer, but nevertheless template strings could be covered here.


getify:
Appreciate the suggestion, and I think you wrote it well.

But the "Up & Going" book, especially ch1, is trying to cover very basic topics at a very high level, as a light easy intro to slowly get into programming, javascript, and the rest of the series.

I've deliberately avoided even mentioning many of the new features in ES6, as I don't want to intimidate right off the bat. Especially since ch1 is not "about JS" per se as much as it is "about programming (w/ JS as example)", I think focusing on a very JS-specific feature would be unnecessarily overloading concepts.

PR #360, opened 2015-03-01
Update ch3.md

pdawyndt:
suggested changes for typo's in "Scope & Closures":"Chapter 3: Function vs. Block Scope"


getify:
Good catches!

PR #361, opened 2015-03-01
Update ch5.md

pdawyndt:
suggested changes for typo's in "Scope & Closures":"Chapter 5: Scope Closures"


getify:
I'll double-check with O'Reilly copy editors, but I'm pretty sure this is correct. Similar to [pluralizing "mother-in-law"](http://english.stackexchange.com/questions/29464/plural-of-daughter-in-law) as "mothers-in-law" instead of "mother-in-laws", I think "passings-around" is correct, not "passing-arounds".

bmacdonald-oreilly:
I'm the one of the O'Reilly editors, and Kyle is correct, for the reason he cites -- it's the noun that gets pluralized. It's definitely one of the fiddly bits of the language, though.

PR #362, opened 2015-03-01
Update apD.md

pdawyndt:
suggested changes for typo's in "Scope & Closures":"Appendix D: Thank You's!"

### DIFF
--- a/scope & closures/apD.md
+++ b/scope & closures/apD.md
@@ -1,11 +1,11 @@
 # You Don't Know JS: Scope & Closures
-# Appendix D: Thank Yous!
+# Appendix D: Thank You's!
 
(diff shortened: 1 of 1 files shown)


scope & closures/apD.md
```diff
@@ -1,11 +1,11 @@
 # You Don't Know JS: Scope & Closures
-# Appendix D: Thank Yous!
+# Appendix D: Thank You's!
```
pdawyndt: Title of the chapter as mentioned in the chapter does not correspond to the title of the chapter as mentioned in the index. So, either one has to be modified.

getify:
Yep, fixing all these to "Acknowledgments" to match the published forms of the books

PR #367, opened 2015-03-01
Update apB.md

pdawyndt:
suggested changes for typo's in "this & Object Prototypes":"Appendix B: Thank You's!"

### DIFF
--- a/this & object prototypes/apB.md
+++ b/this & object prototypes/apB.md
@@ -1,11 +1,11 @@
 # You Don't Know JS: *this* & Object Prototypes
-# Appendix B: Thank Yous!
+# Appendix B: Thank You's!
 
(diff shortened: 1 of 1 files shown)


this & object prototypes/apB.md
```diff
@@ -1,11 +1,11 @@
 # You Don't Know JS: *this* & Object Prototypes
-# Appendix B: Thank Yous!
+# Appendix B: Thank You's!
```
pdawyndt: Title of the chapter as mentioned in the chapter does not correspond to the title of the chapter as mentioned in the index. So, either one has to be modified.

getify:
Yep, fixing all these to "Acknowledgements" to match the published versions.

PR #369, opened 2015-03-01
Update ch3.md

pdawyndt:


getify:
Thanks for all the great fixes! :)

PR #383, opened 2015-03-14
Changed Promise.race([..]) to Promise.race([ .. ]) to be consistent with other usages

ryanplusplus:
Noticed this instance was inconsistent with the style of other instances.


getify:
Thanks!

PR #396, opened 2015-03-18
Fix 'accross'

imurchie:
A simple typo.


getify:
Thanks, good catch! :)

PR #415, opened 2015-04-19
extremely trivial tweak of a comment

JohnTigue:
Just a trivial edit of the output of a console.log( foo, bar) which will not have the comma in the output of the log()


getify:
Thanks! :)

PR #416, opened 2015-04-20
Fix typo for DOM API call `document.getElementById`

booc0mtaco:
Noticed the 'D' in `getElementById` was capitalized. This should be lowercase.


getify:
Thanks!

PR #417, opened 2015-04-21
fix typo ("bascially" to "basically")

mkilics:
fix typo ("bascially" to "basically")


getify:
Just fixed in c556bec8

PR #418, opened 2015-04-21
You Don't Know JS: ES6 & Beyond, Forward

rwaldron:


getify:
There are not enough <3's to throw at this Foreword. I can't even articulate how honored I am. :)

PR #422, opened 2015-04-28
Soft binding code description change

tarnelope:
Based on the provided code, I think we are checking to see if 'this' is 'global' or 'window', instead of 'global' or 'undefined.' 
Please let me know if I'm wrong and why. Thanks.

### DIFF
--- a/this & object prototypes/ch2.md
+++ b/this & object prototypes/ch2.md
@@ -737,7 +737,7 @@ if (!Function.prototype.softBind) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


getify:
There are three conditions which are `||` or'd together:

`!this`, `this === window`, `this === global`

The first one checks a `this` that's `undefined`, `false`, `null`, etc. The second two check if it's the `window` (global obj in the browser) or `global` (global obj in node).

I believe the description as originally written is consistent with this code.

tarnelope:
oh i see! I got tripped up by the !this at the front. Thanks!

getify:
Cheers! :)

PR #429, opened 2015-05-05
Change in indenting.

joeldudley:
V. minimal - just wanted to increase indent of the comment marking the type of StrObject as 'object', so that it lines up with the other comments.


getify:
Thanks!

PR #434, opened 2015-05-07
Fix various errata in Async & Performance Ch. 4: Generators

airandfingers:
I could be wrong/crazy, but isn't this change (062b377) correct?

Step 9. The first request() call should be in the *foo(1) instance, as *foo(2) and *foo(3) are waiting.

Step 10. *foo(2) yields its request promise back to *foo(3), not to *foo(1), which has already completed by this time.


airandfingers:
My second commit (594669c) fixes a clear issue: as written, `thunkify` returns a thunkory that only works the first time it is called. Subsequent attempts will most likely call the callback given to the thunkory in its first call (depending on the implementation of `fn`).

It's worth reviewing, though, as I am changing the way a code example works.

airandfingers:
My third commit (8b5c7df) fixes an obvious typo, a reference to "resolve" that should be "request".

getify:
I can make the change in 8b5c7df as an errata change to the existing first edition. The changes in 594669c need to hold off to second edition. The changes in 062b377 are only partially correct, there are other problems with that list of steps that I need to correct. So I will have to fix that whole list in the second edition.

OahMada:
Why I run the existing `thunkify` code and it works just fine?

getify:
Here's an illustration of how the book's version of `thunkify(..)` utility is "broken":

```js
function thunkify(fn) {
	return function() {
		var args = [].slice.call( arguments );
		return function(cb) {
			args.push( cb );
			return fn.apply( null, args );
		};
	};
}

var f = thunkify((x,y,cb) => cb(x + y));

var seven = f(3,4);

seven(sum => console.log(`Seven: ${sum}`));    // Seven: 7
seven(sum => console.log(`Another Seven: ${sum}`));    // Seven: 7     <---- Oops
seven();   // Seven: 7    <------- WHAT!?
```

We would have expected the next-to-last line to print `Another Seven: 7`, but it didn't. The shared closure over `args` modified the list in the first `seven(..)` call to add the first version of the callback. From then on, all subsequent calls to `seven(..)` don't pay any attention to what callback you pass in.

Compare to the *corrected* `thunkify(..)`:

```js
function thunkify(fn) {
	return function() {
		var args = arguments;
		return function(cb) {
			var thunk_args = [].slice.call( args );
			thunk_args.push( cb );
			return fn.apply( null, thunk_args );
		};
	};
}

var f = thunkify((x,y,cb) => cb(x + y));

var seven = f(3,4);

seven(sum => console.log(`Seven: ${sum}`));     // Seven: 7
seven(sum => console.log(`Another Seven: ${sum}`));     // Another Seven: 7
```

So, yes, the book's version is broken and that's regrettable. But it was enough of a substantive change (and would have needed corrective text to explain), and the book had already been out for quite awhile at that point... so I opted not to modify the book. At some point, a second edition will be written, and this issue will be addressed then.

PR #439, opened 2015-05-08
es6+beyond: revise const GC admonition

### LINKED ISSUE #437: es6: ch2: "const declarations", GC admonition may be inaccurate
Per [the text](https://github.com/getify/You-Dont-Know-JS/blob/e49c289219dd88e9552c9c3de0c3174af6e38554/es6%20%26%20beyond/ch2.md#const-declarations):

> **Warning:** Assigning an object or array as a constant means that value will never be able to be garbage collected, since the reference to the value can never be unset. That may be desirable, but be careful if it's not your intent!

I don't believe this is so – GC works as usual on objects pointed to by these references. If they can't be reached by walking the graph from a retaining object, they'll be GC'd like any other object. The distinction is that global constant values will *always* be reachable from a retaining object – the global object – so constants at that level may never be garbage collected.

As an example, just to illustrate:

```javascript
// will never be GC'd in this program, since each of these create
// a reference reachable from the global object:
var globalValue = {}
const constantGlobalValue = {}
let blockScopedGlobalValue = {}

{
  const someOtherObject = {}
}
// at this point someOtherObject is a viable candidate for GC, since it
// can no longer be reached by a retaining object.
```

chrisdickinson:
Fixes #437.

### DIFF
--- a/es6 & beyond/ch2.md
+++ b/es6 & beyond/ch2.md
@@ -205,7 +205,7 @@ You are not allowed to change the value of the variable once it's been set, at d
 
 The `a` variable doesn't actually hold a *constant* array, it holds a *constant* reference to the array; the array itself is freely mutable.
 
-**Warning:** Assigning an object or array as a constant means that value will never be able to be garbage collected, since the reference to the value can never be unset. That may be desirable, but be careful if it's not your intent!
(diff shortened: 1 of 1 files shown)


getify:
LGTM. :)

PR #440, opened 2015-05-09
Fix typo from `he` to `it`

neu5:
Change the type of entity from male to neuter because it's about the `function`


getify:
Good catch! :)

PR #444, opened 2015-05-11
Combine duplicate mentions of jsperf setup/teardown code

airandfingers:
Lines 199 and 203 both describe `setup` and `teardown` code for jsperf tests. This is duplicate information, and one should be removed.

In my opinion, it makes more sense to remove the mention on line 199, as the **Note** on line 201 expands on the first sentence of line 199.


getify:
The second mention as originally written is only slightly repetitive and is done for the purposes of connecting in the readers' minds that the jsperf setup/teardown settings map to the benchmark.js setup/teardown config. I don't see that there's any need for change here.

getify:
**Note:** I might slightly clean up the second reference's wording in the *second edition*, but there's no need for change as errata.

PR #445, opened 2015-05-12
Fixed a typo

zudov:
This `ol'` word clearly looks as a typo. I am not sure if `old` was intended meaning, but it seem to match. Correct me if I guessed it wrong.

### DIFF
--- a/this & object prototypes/ch2.md
+++ b/this & object prototypes/ch2.md
@@ -397,7 +397,7 @@ For example, the `Number(..)` function acting as a constructor, quoting from the
 >
 > When Number is called as part of a new expression it is a constructor: it initialises the newly created object.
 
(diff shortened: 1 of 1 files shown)


zudov:
If it's not a typo and `ol'` carries some meaning it probably worth to explain it.

getify:
it's American slang to abbreviate **old** to **ol'**, mostly for a folksy or informal tone. It's often used in phrases like "good ol' time" or "my ol' man" (my father).

Agreed it's not the easiest to google search for unless you know what to look for already. Not sure it matters too much. But saying "any old function" sounds wrong to American *ears*, so if it's going to be fixed, the word should just be dropped altogether.

zudov:
@getify Oh. I see, it's my confusion after all. Thanks for explanation.

Considering this detail I think it would be better to leave it as it is.

getify:
Cheers!

PR #457, opened 2015-05-29
Update ch5.md Map api // includes -> has

DJTB:
I've looked through the documentation and can't see ```.includes(..)``` ever working with maps. The polyfill for ```.includes(..)``` uses ```.length``` (not ```.size```) so perhaps that's why they opted for 'has' rather than 'includes'. 

Seems ```.includes(..)``` is only for use with strings/arrays/typed arrays.

### DIFF
--- a/es6 & beyond/ch5.md
+++ b/es6 & beyond/ch5.md
@@ -300,11 +300,11 @@ var vals = [ ...m.values() ];
 vals;							// ["foo","bar"]
 ```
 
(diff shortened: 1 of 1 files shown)


DJTB:
Although it's also very possible your example was intended to be:
``` js
vals.includes( "foo" );
vals.includes( "baz" );
```
Since you did ```var vals = [ ...m.values() ];``` just above your sentence about .includes(..)

DJTB:
```.has()``` only works for Map keys, my bad. Have put through another pull request with an updated fix.

PR #458, opened 2015-05-29
Update ch5.md Map values .includes() example

DJTB:
As far as I can tell you intended to use ```includes(..)``` on the array ```vals``` gathered from ```m.values()```.
Using ```includes()``` on the Map itself will throw a TypeError.

### DIFF
--- a/es6 & beyond/ch5.md
+++ b/es6 & beyond/ch5.md
@@ -303,8 +303,8 @@ vals;							// ["foo","bar"]
 To determine if a value exists in a map, use the `includes(..)` method (which is the same as on standard arrays as of ES6):
 
 ```js
-m.includes( "foo" );			// true
-m.includes( "baz" );			// false
+vals.includes( "foo" );			// true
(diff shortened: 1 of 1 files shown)


getify:
Hmm... I need to figure out what I was thinking here. There's no `Array#includes(..)` yet (it's [proposed for ES7](https://github.com/tc39/Array.prototype.includes)). I don't know where I got the idea that it was in ES6, nor where I got the idea it was (or was going to be) on `Map`.

The only way I can see this working right now in ES6 is `vals.find(..)` (or `vals.indexOf(..)` of course). That's certainly not what I intended to get across. So I've clearly just made a mistake.

I might just need to take out that part of the section entirely. I'll research a bit more. Thanks for bringing it to my attention.

DJTB:
Oh you're right it is ES7, I was testing things in babel(-node) and assumed it was ES6 as well. Just following along with the book and trying things out. Thanks for the incredible resource!

PR #461, opened 2015-06-05
Update ch5.md

amirjalilifard:
I have read this document various times. Using "a" as the name of the new instance of object "Foo" is very confusing. The new object's name can be "Foo2" or other name related to "Foo" function. So I suggest using "Foo2". I believe every other names is better that "a" since sometimes it is very confusing and reader has to return and read the previous paragraph and code peaces again and again.

### DIFF
--- a/this & object prototypes/ch5.md
+++ b/this & object prototypes/ch5.md
@@ -170,12 +170,12 @@ function Foo() {
 	// ...
 }
 
-var a = new Foo();
+var Foo2 = new Foo();
 
 Object.getPrototypeOf( a ) === Foo.prototype; // true
 ```
 
-When `a` is created by calling `new Foo()`, one of the things (see Chapter 2 for all *four* steps) that happens is that `a` gets an internal `[[Prototype]]` link to the object that `Foo.prototype` is pointing at.
+When `Foo2` is created by calling `new Foo()`, one of the things (see Chapter 2 for all *four* steps) that happens is that `Foo2` gets an internal `[[Prototype]]` link to the object that `Foo.prototype` is pointing at.
 
 Stop for a moment and ponder the implications of that statement.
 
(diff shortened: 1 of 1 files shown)


amirjalilifard:
Hey guys;
I suggest using another name for "a" and "b" as new Instances of Object "Foo" since it is confusing. I asked several readers about this part and almost everybody thinks "a" is a confusing name.

getify:
I can appreciate the concerns, but the suggestion you're making makes these parts of code inconsistent with every other chapter and book in the series, where I always use variable names like `a`, etc. I am not anytime soon going to update every single code snippet across all six books, and consistency is important to me.

Just as a quick explanation: I use generic variable names because of *cognitive load theory*, where the concern is that if I use real variable names in real software contexts, then the reader is trying not only to learn the programming concept I'm teaching, but also to learn the software context (if they're not already familiar with it). Turns out there's just not that many universal contexts, so any one I pick will work great for some and much worse for others. So the happy medium is to use generic names to reduce cognitive load across the broadest range of audience.

amirjalilifard:
Got it. @getify I wanna participate in developing this book. Is there any special part I can take part in? If there is, please tell me. I think I can be an efficient contributor.

getify:
@amirjalilifard

Appreciate the volunteering. If you look around at the content, you'll see that all six books in the series are complete now (been in progress the last 18-24 months). Well, the first edition is done, anyway. All six have been published by O'Reilly, too (the last one will be on shelves soon). So there's really nothing left to do on first edition.

So as for helping, I believe the most assistance will be had at a later time as the second edition edits start happening. Of course, I'll happily accept any editorial comments for consideration at any time, but I won't be actively working on the second editions until probably this fall at the earliest. Keep your eyes on the repo, and thanks in advance for any future collaborations. :)

PR #466, opened 2015-06-12
Fix 'writable: false' to 'writable: true'

breda:
At line 97, this :
**and it's not marked as read-only (`writable:false`)**
should actually be this : 
**and it's not marked as read-only (`writable:true`)**

### DIFF
--- a/this & object prototypes/ch5.md
+++ b/this & object prototypes/ch5.md
@@ -94,7 +94,7 @@ If the property name `foo` ends up both on `myObject` itself and at a higher lev
 
 As we just hinted, shadowing `foo` on `myObject` is not as simple as it may seem. We will now examine three scenarios for the `myObject.foo = "bar"` assignment when `foo` is **not** already on `myObject` directly, but **is** at a higher level of `myObject`'s `[[Prototype]]` chain:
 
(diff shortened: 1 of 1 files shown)


getify:
Please see:

https://github.com/getify/You-Dont-Know-JS/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+writable

breda:
Ah! Sorry about that @getify 

But, as you stated, that the (`writable:false`) is only associated with the word "**read-only**",
then I suggest adding something to clarify that, leaving it like this will probably confuse people, and/or lead to more pull-requests like this one.

getify:
I am aware of the running confusion. Paradoxically, the more this issue comes up, and I have to rethink through my reasoning, the more I get convinced the original wording was correct and I'm less understanding of the other interpretation. What, grammatically, leads people to assume the parenthetical applies to the whole phrase and not just the term it immediately proceeds?

That having been said, it's already definitely something that will be addressed in the second edition. I'll use your bug specifically for tracking that. :)

guser9822:
@getify Probably people get confused by this allegedly typo because the first sentence is in bold, differently from the second sentence. But, honestly, reading the second sentence after the first, I get confused too, and I convice myself that it's a typo

PR #467, opened 2015-06-14
Corrects minor typo.

michaelck:
Omits the first occurrence of "be" from the following sentence on line 580: "**Note:** Separate files **"foo.js"** and **"bar.js"** would be need to be created..."


getify:
Thanks! :)

PR #479, opened 2015-07-05
Update ch3.md - remove unnecessary "the"

cardern:
Thanks for writing this excellent book. There is a minor typo in the linked section. "The" is unnecessary before "Chapter 5".

"We'll come back to the important differences between these two operations in the Chapter 5 when we explore [[Prototype]]s in detail."

https://github.com/getify/You-Dont-Know-JS/blob/master/this%20%26%20object%20prototypes/ch3.md#existence


getify:
Thanks!

PR #481, opened 2015-07-06
s/positive/non-negative/

masak:
Zero (which is also given as an example) is allowed as an index, but not a positive integer.


getify:
Good suggestion.

PR #483, opened 2015-07-06
fixed typo

matchabros:
Just a quick typo fix.

### DIFF
--- a/this & object prototypes/ch3.md
+++ b/this & object prototypes/ch3.md
@@ -461,7 +461,7 @@ The final `defineProperty(..)` call results in a TypeError, regardless of `stric
 
 **Note:** There's a nuanced exception to be aware of: even if the property is already `configurable:false`, `writable` can always be changed from `true` to `false` without error, but not back to `true` if already `false`.
 
-Another thing `configurable:false` prevents its ability to use the `delete` operator to remove an existing property.
+Also, configurable:false prevents you from using the delete operator to remove an existing property.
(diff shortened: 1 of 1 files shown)


getify:
That change makes the sentence into a fragment, since it removes the verb "is".

matchabros:
But it sounds so weird! Anyway, apologies for the unnecessary commit.

getify:
Could perhaps have had a comma or colon in between "is" and "the", such as:

> Another thing `configurable:false` prevents is, the ability to use the `delete` operator to remove an existing property.

> Another thing `configurable:false` prevents is: the ability to use the `delete` operator to remove an existing property.

Do either of those changes make it read clearer to you?

bmacdonald-oreilly:
From a grammatical standpoint, there shouldn't be any punctuation there. However, when the "sounds wrong" objection comes up, the solution is usually "recast the sentence," like so:

configurable:false also prevents you from using the delete operator to remove an existing property.

Or if you don't like starting the sentence with a lower-case code term:

Also, configurable:false prevents you from using the delete operator to remove an existing property.

matchabros:
Changes proposed by bmacdonald-oreilly are fine. 

Don't get me wrong, everything is clear in that sentence, but feels weird while reading it - knocks you out of the flow to be more precise.

shivaylamba (changes requested):
Please fix merge conflicts.

PR #484, opened 2015-07-09
Update ch2.md

stebru:
Make "for" condition evaluate to true.


getify:
Good catch this was a typo from combining code from another snippet. However, the best fix is to remove the `!`, not add another one unnecessarily. :)

stebru:
Right, cool :)

PR #489, opened 2015-07-21
Correct the result of code snippet

hbkdsm:
The snippet just simply logs `undefined` to the console, rather than throws a ReferenceError. In your case, both `this` keyword in function `foo` and `bar` refers to the global object, namely `window` in non-strict mode. Since no variable named `a` is defined in global scope, `console.log(this.a)` will logs `undefined`!

### DIFF
--- a/this & object prototypes/ch1.md
+++ b/this & object prototypes/ch1.md
@@ -251,7 +251,7 @@ function bar() {
 	console.log( this.a );
 }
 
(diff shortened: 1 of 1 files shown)


hbkdsm:
When you access a property of object, you will never get a ReferenceError.

getify:
Thanks good catch

PR #492, opened 2015-07-29
correct word spelling: presize->resize

creeperyang:


getify:
"presize" is the intended wording here. It's intended to suggest "pre-sizing", as in setting up the size of the array before filling it with contents.

PR #495, opened 2015-07-30
instructional comment corrected

pelonpelon:
Technically this comment:
// hard-bound `bar` can no longer have its `this` overridden
is incorrect.
`bar` can most definitely have its `this` overridden as the following code demonstrates:

```javascript

function foo() {
    console.log( this.a );
}

var obj = {
    a: 2
};

var obj2 = {
    a: 42
};

var bar = function() {
    foo.call( obj );
    console.log(this.a)
};

bar(); // 2
setTimeout( bar, 3000 ); // 2

// `foo` is hard-bound to `obj` inside `bar` and can no longer have its `this` overridden
// but `this` inside bar can be overridden without affecting `this` of `foo`
 
bar.call( obj2 ); // 2 // 42


```

### DIFF
--- a/this & object prototypes/ch2.md
+++ b/this & object prototypes/ch2.md
@@ -283,7 +283,7 @@ var bar = function() {
 bar(); // 2
 setTimeout( bar, 100 ); // 2
 
-// hard-bound `bar` can no longer have its `this` overridden
+// inside `bar`, `foo` is hard-bound to `obj` and can no longer have its `this` overridden
 bar.call( window ); // 2
 ```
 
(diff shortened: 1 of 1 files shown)


pelonpelon:
The closer I read your prose, I wonder if you might be saying that `bar` is hard-bound. If that's true, I'm not following very well, it appears to me that `foo` is hard-bound. Regardless, `bar` can still have its `this` overridden. Maybe you have a better re-wording in mind.

getify:
Hopefully that change clarifies the comment.

pelonpelon:
Yeah, thanks.
I actually closed the PR because I realized I couldn't follow my own logic. I've been faking my understanding of `this` for years. I understand it infinitely better now that I've puzzled over your excellent book!

getify:
Cheers! :)

PR #497, opened 2015-07-31
minor edits: verb agreement, temporal consistency

pelonpelon:
in line 445 it's difficult to see, but I've changed "**But, it used**" to "**But, it uses**"


getify:
Thanks

PR #505, opened 2015-08-06
join JSON array elements with commas

masak:
This comma is indeed never used (as that is the point of the example), but
the original intent was to (mistakenly) render a JSON string, and a JSON
string would join the values with a comma.

### DIFF
--- a/types & grammar/ch4.md
+++ b/types & grammar/ch4.md
@@ -166,7 +166,7 @@ var b = {
 	// probably incorrect!
 	toJSON: function(){
 		return "[" +
-			this.val.slice( 1 ).join() +
+			this.val.slice( 1 ).join(", ") +
 		"]";
 	}
 };


getify:
As [shown here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join), the default behavior of `join()` with no arguments is to assume `","` as the joiner.

> This comma is indeed never used (as that is the point of the example)

Actually, it *is* used, since the call is `this.val.slice(1)`, which would be the array `[2,3]`. That's why I set up the example that way. :)

masak:
Wow, live and learn. I had no idea about the `separator = ','` default. And indeed, I also misread where the stringification happens.

The book series keeps living up to the promise of its name. ☺ Closing pull request.

getify:
Cheers. :)

PR #508, opened 2015-08-07
fix two s/equitable/equatable/ typos

masak:
I think what happened here is that some over-eager spell-checker liked "equitable" better.
But "equitable" seems to mean "characterized by equity or fairness; just and right; fair;
reasonable", and "equatable" (as in "coming up as equal in an equality comparison") makes
a lot more sense here. A third "equatable" earlier in the first paragraph seems to support
this, too.


getify:
Thanks, great catch!

PR #511, opened 2015-08-09
fixed logical mistake in explanation of iterator loop

Mobiletainment:
The description states, that `res.done` is consulted in each iteration of the `for equivalent of a for..of loop` example and if res.done is `false`, the iteration doesn't occur. In fact, it's the other way around: if `res.done` is `true`, the statement evaluates to false and the iteration doesn't occur.

### DIFF
--- a/es6 & beyond/ch3.md
+++ b/es6 & beyond/ch3.md
@@ -174,7 +174,7 @@ for (var v, res; (res = it.next()) && !res.done; ) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


getify:
Would you mind updating the word "statement" to "expression" in your PR? Other than that small tweak, your update is great. Thanks!

getify:
actually, i see this is already part of another PR. i'll address it there.

PR #517, opened 2015-08-23
Fix Typo

th11:


getify:
Hmm... @bmacdonald-oreilly @jtassia, what do you think here? Is this a miss from copyedit?

jtassia:
Oops... sorry I missed this, Kyle!

How about one of these options instead?

"...despite clearly allowing you to write asynchronous JS code..."

OR

"...despite clearly allowing asynchronous JS code..."

PR #522, opened 2015-08-26
Inconsistencies around setTimeout invocation

zackgao:


getify:
my standard style does not have a space before a closing ). if that happened somewhere, it's a minor formatting error. but i don't really see any need to tweak these as provided, since it's not according to the normal style throughout.

PR #527, opened 2015-08-30
fix statement about weak typing

mzahor:
Weak typing is comparable to strong typing, not to static one. Language can be weakly and statically typed at the same time.

http://stackoverflow.com/questions/2351190/static-dynamic-vs-strong-weak:

> Static/Dynamic typing is about when type information is aquired (either at compile time or at runtime)
> Strong/Weak typing is about how strictly types are distinguished (e.g. whether the language tries to do implicit conversion from strings to numbers).

### DIFF
--- a/up & going/ch1.md
+++ b/up & going/ch1.md
@@ -343,7 +343,7 @@ The easiest way to go about that in your program is to assign a value to a symbo
 
 In some programming languages, you declare a variable (container) to hold a specific type of value, such as `number` or `string`. *Static typing*, otherwise known as *type enforcement*, is typically cited as a benefit for program correctness by preventing unintended value conversions.
 
-Other languages emphasize types for values instead of variables. *Weak typing*, otherwise known as *dynamic typing*, allows a variable to hold any type of value at any time. It's typically cited as a benefit for program flexibility by allowing a single variable to represent a value no matter what type form that value may take at any given moment in the program's logic flow.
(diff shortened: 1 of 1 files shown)


getify:
I don't think those assertions match with what's indicated here:

https://en.wikipedia.org/wiki/Strong_and_weak_typing

mzahor:
This topic is a bit controversial, but I would rather believe PhD in programming languages (Dan Grossman, Programming Languages course on Coursera [youtube lesson](https://www.youtube.com/watch?v=SA6Nbp0Fl5g)) than Wikipedia. If you still not convinced, just abandon the PR, I'm absolutely ok with it, don't want to start a flame on this topic.

getify:
I appreciate the spirit of trying to make the books as accurate as possible. And often I'm quite a stickler on words and phrases and their intended meanings. It regularly frustrates me when common usage bastardizes the original intent.

OTOH, this particular topic is wrought with dissension and nuance, as you point out. Combine that with the fact that this is the first beginner book in the series, which is intended for those with little to no experience whatsoever. My goal was to use things not just accurately but also in a way that won't actively confuse or overwhelm.

To explain why these phrases are so complex to get precise definitions for, and to give it the full attention it deserves, would be wholly inappropriate for a beginner book. Those are the kinds of things that PhD's should write about (of which I am not, and certainly that's not the intent of this book or series).

So, while I appreciate your sincere effort, and I can understand where you're coming from, I don't see it as very useful to change the wording to things which won't match with what a beginner reader would get if they google searched (and found wikipedia). And since I can't fully explain all the nuance, I'm left feeling the best approach is the current one.

No flame war intended or taken here. Thank you for your contribution. It's very appreciated.

PR #536, opened 2015-09-14
Fix typo about failed LHS lookup error type in Review section

eduard-sukharev:
I believe it's a typo, and quite a misleading one, as Review section differs from what's actually written exactly before it.


getify:
I don't believe this is accurate. Both unfulfilled RHS and unfulfilled LHS (in strict mode) result in `ReferenceError` errors, as indicated in the text. This is verifiable:

lhs-rhs

What part do you think this contradicts "exactly before it"?

eduard-sukharev:
Sorry, you're right - my bad. For me as a learner it was a bit distracting that you write about two types of errors, but then mentions only one of them in the review section.

Sorry again, closing this PR and going off to read your books :)

getify:
No worries. :)

PR #542, opened 2015-09-22
Added link for purchase/download of the first book

elahmo:
Other books have working link, this one doesn't have it.


getify:
Good catch, thanks!

PR #544, opened 2015-09-24
Fix typo in ch8: "signficantly" => "significantly"

jabbrass:
Typo in Es6&beyond, ch8, line 403: "signficantly" => "significantly".
Thanks for the thorough material, best reference for JS out there!


getify:
Thanks!

PR #546, opened 2015-09-26
Update to preface.md -- more colorful analogy

mikece:
With a nod to Mark Twain, who said the difference between the right word and almost the right word is the difference between lightning and a lightning bug, I submit a more colorful analogy -- one I've used for years -- to underscore the difference between the languages of Java and JavaScript.


getify:
I appreciate the suggestion.

There are many, many analogies used for this topic. I chose one that I like. Many others use different ones. I'm not sure your analogy offers any more weight (unless the reader gets the reference as a Mark Twain fan), so I'll defer changing things for now, but I'll at least consider it in the future for second edition updates.

getify:
Closing as not applicable anymore, since the 2nd-ed books have a rewritten preface.

PR #547, opened 2015-09-28
Fix typo in ch3 (es6 & beyond)

jabbrass:
Typo on line 1476: "additionaly" => "additionally". Thanks again for the great material.


getify:
Thanks!

PR #549, opened 2015-09-30
ACCESSORY_PRICE conditional logic leads to overdraft

rystan:
The conditional to check whether or not one can afford the additional accessory should include the accessory price, otherwise if the bank_balance was $100.00, the code would determine "I'll take the accessory" even though the new total amount is more than their bank balance.


getify:
Closing as not applicable anymore, since the new book "Get Started" (2nd-ed) has replaced "Up & Going".

PR #553, opened 2015-10-07
Fix typos in ch4 (async & performance)

jabbrass:
Line 760: "capabililty" => "capability"
Line 2178: "unitialized" => "uninitialized"


jabbrass:
Thanks again for the fantastic material! I hope you don't mind the rather numerous, but insignificant typo corrections...

getify:
Thanks! Most of these are already caught by the copy-editors in the published editions, but I appreciate having the fixes ported to the repo too. :)

PR #555, opened 2015-10-12
Remove broken link

fix-fix:
Remove link to "Don't Release Zalgo!" from "Async & Performance: Chapter 2: Callbacks",
because original link is broken and that post cannot be found anywhere on the Internet


getify:
Fixed better by #596.

PR #558, opened 2015-10-12
Match ES6 Review Section Title to Other Books

benbrandt:
Found the same issue as #556 and #557 in the ES6 & Beyond book.

The Review section titles were missing the (TL;DR) found in previous
books in the series.


bmacdonald-oreilly:
We left the (TL;DR) out of the print versions of the books, because Kyle's editor (that'd be me) is an old man with no sense of humor. That's why it doesn't appear in the later titles.

benbrandt:
Oh got it! I'll go ahead and close my three pull requests then. Sorry about that!

bmacdonald-oreilly:
No worries. As an editor, I'm a big admirer of consistency. Thanks for pointing it out!

getify:
Yeah, these are things that need to get cleaned up with the second editions, too. So that the repo more closely matches what's ending up in the print books. :)

PR #560, opened 2015-10-13
Fix Typo in Ch 2 of Types & Grammer

benbrandt:
I am not an English major, but it seemed like this sentence was missing a “to”. Let me know if it should be written another way. Thanks!


getify:
Thanks!

PR #562, opened 2015-10-15
Add new phone only if bank_balance can cover the cost

michelrossier:
CH1, Exercise: Add new phone only if bank_balance can cover the cost


getify:
will be covered by #549

PR #565, opened 2015-10-20
Update ch6.md

privatmamtora:
Tested in chrome if bug still exists, it doesn't


privatmamtora:
Either add a note or remove whole `Debugged` section

PR #570, opened 2015-10-22
Fix typo ch7 (es6 & beyond)

jabbrass:
Line 884: "properities" => "properties"
Thanks again for your superb material!


getify:
Thanks!

PR #571, opened 2015-10-23
Fix typo

li-xinyang:
Move comma outside the quotation marks.

### DIFF
--- a/up & going/ch1.md
+++ b/up & going/ch1.md
@@ -152,7 +152,7 @@ This should look similar to the following:
 
 <img src="fig2.png" width="500">
 
-Once you submit the input text by clicking "OK," you'll observe that the value you typed is stored in the `age` variable, which we then *output* with `console.log(..)`:
+Once you submit the input text by clicking "OK", you'll observe that the value you typed is stored in the `age` variable, which we then *output* with `console.log(..)`:
 
 <img src="fig3.png" width="500">
 


getify:
I think the current version is consistent with the style used in the book copy-editing. @bmacdonald-oreilly

bmacdonald-oreilly:
Kyle is correct; punctuation inside the quotes is consistent with our house style.

li-xinyang:
May I know the reason why the punctuation should be inside the quotes? Please ignore me if the question is too silly to answer.

bmacdonald-oreilly:
In American English, commas and periods go inside quotation marks. Chicago Manual of Style, 16th Edition, section 6.9: "Periods and commas precede closing quotation marks, whether double or single." 

That's pretty clear, but with all most questions of grammar, there's some wiggle room. British English often (but not always) puts the punctuation outside the quotation mark. You could also argue that because we're quoting text from the UI, and there's no comma on the OK button in the UI, it shouldn't be inside the quotes. Finally, some copyeditors consider the quotation marks optional in this case. If I say "Click the OK button," it's pretty obvious that I mean the button marked "OK," even without the quotes. But if I say "Click the Click here to go on link," that's pretty confusing. Saying "Click the 'Click here to go on' link" is better, although "Click the link that says 'Click here to go on'" is even better.

getify:
Thanks for the clarifications!

li-xinyang:
Thank you so much for the clarification!

PR #574, opened 2015-10-24
simple fix for a not really module

brescia123:
An object with a function property on it alone is not *really* a module. Fixing a snippet based on this assumption.


getify:
This is already a "module" because it closes over the `id` variable (parameter). Also, your change only introduces an unused local variable -- that doesn't really do anything substantive.

PR #578, opened 2015-10-30
Removes unnecessary hyphen in reasonable

melindabernrdo:


getify:
Thanks, but the hyphen is intentional. it's signifying the difference between "reasonable" (something that's ok and acceptable) with something that's "reason-able" (something that's able to be reasoned about).

PR #580, opened 2015-11-03
Fix spelling typo

fay-jai:
The only change I made was fixing the spelling typo:
1) Change from that to that's
No other changes made.


getify:
Thanks!

PR #583, opened 2015-11-07
Async & Performance, Ch 2: `timeoutify()` example bug

btoll:
The example will always log to `console.error`.

A null value must be pushed onto the front of the list in the success case so it will fail
the `err` clause in the callback.


getify:
Good catch, thanks.

PR #589, opened 2015-11-23
(fix) change "which" to "that"

mmurkidjanian:
This fix alters one word that potentially causes misinterpretation. The first time I read this sentence, I thought it meant: "The omission of a name is often helpful in providing more readable/understandable code." I quickly realized that this is not the case, and I believe this potential confusion can be avoided by changing "which" to "that."


getify:
Thanks!

PR #592, opened 2015-11-25
Fix typo Async & Perf - Ch6 - Context Is King

masonlouchart:


getify:
**NOTE:** I need to just swap the 80 and 100 in the X and Y... because 100 is not 80% of 120.

LM450N:
@getify I did the changes (the swap). Sorry for my first bad proposal.

getify:
thanks! :)

PR #593, opened 2015-11-26
Typo correction

alisterlf:
Changing
this
(`a > 41` here)
to this
(`a > 41`) here

### DIFF
--- a/up & going/ch2.md
+++ b/up & going/ch2.md
@@ -512,7 +512,7 @@ var b = (a > 41) ? "hello" : "world";
 // }
 ```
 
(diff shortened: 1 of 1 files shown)


getify:
I don't think your change makes the text more clear, but instead I think it ever so slightly makes it less clear. Can you explain your reasoning?

huei90:
Why don't just remove the word `here`?

alisterlf:
When I read it, I thought it was is a typo the `here` being inside the expression.
For this reason I just moved the word out of expression.
But I agree with @huei90 , the word "here" can be removed.

getify:
Doesn't seem like a justified change, so closing for now. If there's further discussion to be had, please feel free to re-address.

PR #598, opened 2015-12-04
(fix): `there's` to `there're` in async, ch.2

ggauravr:


getify:
The O'Reilly copyeditor didn't make this change in the published form. I also wouldn't do something strange like `'re` contraction, so if it needed to be changed to the plural form, I would have un-contracted it to "there are".

PR #599, opened 2015-12-07
correct various grammatical errors

mmurkidjanian:
This commit fixes various syntactic errors found in Chapter 3 of the book *this & Object Prototypes*. The corrections enhance readability.

### DIFF
--- a/this & object prototypes/ch3.md
+++ b/this & object prototypes/ch3.md
@@ -162,9 +162,9 @@ myObject["[object Object]"];	// "baz"
 
 ### Computed Property Names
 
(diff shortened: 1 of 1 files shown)


getify:
some of your changes (like the commas) are ones I would pull in as trivial. Others represent wording changes, and the copy editor for O'Reilly handles those. It is possible that some of the copyedits from the published book didn't make it back into this repo, but I want to avoid actual wording changes here that are not in the published form. At best, some of these can be considered decent suggestions for second edition revisions, so I'll mark it with that label to come back to later.

PR #603, opened 2015-12-14
Add: note about type names

elkebirmed:
I think this will help JS developers from getting false when they're expecting true

### DIFF
--- a/types & grammar/ch1.md
+++ b/types & grammar/ch1.md
@@ -60,7 +60,7 @@ typeof { life: 42 }  === "object";    // true
 typeof Symbol()      === "symbol";    // true
 ```
 
-**Note:** Type names written in lowercase. `string` is the type name of the `String` object.
+**Note:** the string values returned from `typeof` are all lowercase.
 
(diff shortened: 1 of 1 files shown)


getify:
I think this is inaccurate. `String` is an `object`, that is boxed (wrapped) around a `string` primitive. They're correlated, but not at all the same thing.

elkebirmed:
That's right but let's see this example:

```javascript
typeof 'Hello' === 'string' // true
typeof 'Hello' === 'String' // false (this typo will change the world :D)
```

did you get what i mean, JS developers will think that the type is not a string while it's so.

getify:
The note could say "Note: the string values returned from `typeof` are all lowercase." That doesn't help that much, but maybe just a slight bit.

But as written, I think the current PR is inaccurate/misleading. If you'd like to update the PR to the above suggested wording, I could consider that for *second edition*.

elkebirmed:
Does this help!

getify:
Thanks.

PR #604, opened 2015-12-14
Patch

Projjol:
Improves navigation


getify:
This repo serves the primary purpose of providing source documents to the editor (O'Reilly) for publishing. That's why I don't add flourishes like extra navigation into the main chapter docs, because that stuff would be noise they'd have to remove during import.

PR #605, opened 2015-12-18
some elaboration.

31H0B1eV:
Just trying it in chrome console and have a misunderstanding with behavior.


getify:
Your code example is missing a `var` so it's creating an auto-global (a big no-no). Even still, in a fresh environment, the assignment will definitely not have hoisted. See:

screen shot 2016-01-15 at 7 59 10 am

PR #607, opened 2015-12-18
Corrected `writable:false` to `writable:true` for non-readonly objects

schatekar:
The sentence refers to an object that is *not* marked as read-only but then indicates that with `writable:false` which is not correct. Fixed it to be `writable:true`


getify:
Duplicate. Please see: https://github.com/getify/You-Dont-Know-JS/pull/466

PR #608, opened 2015-12-18
explain why or how instead of what

awolfe76:
Simple update to have comment tell "why" instead of "what".

> Comments should explain *why*, not *what*. They can optionally explain *how* if that's particularly confusing.


getify:
You're absolutely right under normal circumstances. I broke my own rules here because I'm using code comments as inline book commentary to describe the *what* as I teach it. Good thinking here, but I did it on purpose. "You first have to know the rules before you can break them." :)

PR #614, opened 2015-12-24
(fix): corrected values in sample output

corinna000:
value of a should be 11, value of b should be 31. See http://plnkr.co/edit/ZtTx3Uc1uiIEfXf6cwJs?p=preview


getify:
The comments on those two lines show the outputs of the `console.log(..)` statements, not the values of `a` and `b` respectively. In fact, deliberately, they show the `b` and `a` in reverse order, showing when the side effects are occurring.

corinna000:
Okay, I get it now. I'm really used to seeing that // syntax as being used to explain the value of the variable that's being assigned on that line.

PR #619, opened 2016-01-01
Reworded so function timer is not referred to as anonymous

aquascaper:
I don't think timer is an anonymous function in this case since it has a name.


getify:
Good fix, thanks!

PR #622, opened 2016-01-03
Change the copyright to 2016

kmoretz:
since it's 2016 now this should probably be done...


getify:
The copyrights will be updated for second edition, later this year. Thanks.

PR #623, opened 2016-01-04
Replace `:` with `,`.

aruprakshit:
I think `:` was a typo there.


getify:
Nope, not a mistake. The `"instantiation", "inheritance" and "(relative) polymorphism"` are three kinds of the `mechanics of "classes"` to be discussed. :)

aruprakshit:
@getify  Sorry for getting it wrong. :)

PR #626, opened 2016-01-08
Remove extraneous word "property"

msach:
It seems to me that the word "property" doesn't need to be at both ends of the `Foo.prototype` text. Leaving the second one flows better when reading, I think.


getify:
Thanks! :)

PR #627, opened 2016-01-08
It's safety to use loose equality.

wKich:
Based on chapter 4 of Types & Grammar.

### DIFF
--- a/async & performance/ch3.md
+++ b/async & performance/ch3.md
@@ -372,12 +372,12 @@ The general term for "type checks" that make assumptions about a value's "type" 
 
 ```js
 if (
-	p !== null &&
+	p != null &&
 	(
-		typeof p === "object" ||
-		typeof p === "function"
+		typeof p == "object" ||
+		typeof p == "function"
 	) &&
-	typeof p.then === "function"
+	typeof p.then == "function"
 ) {
 	// assume it's a thenable!
 }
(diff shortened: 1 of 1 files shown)


getify:
eh. i don't see this change adding much value, as this book is not dealing with the topic of coercion.

wKich:
But, I think it's good point to use your own opinion in code examples. Instead, why you talking about coercion and code simplicity in the one book, and not use it in all other circumstances?

getify:
In particular, the objectionable part of the change to me is the `p1 == p2`, which is a comparison of references. The use of `==` here implies that coercion is OK (and perhaps even expected), which is definitely not the case for comparison of references.

The point of using coercion is to make code clearer, not for consistency of belief. In fact, the point of all coding is for cleaner communication of intent.

I don't think that saving the `=` in these cases is buying much in terms of readability, even if you're exactly right that it's most consistent with the arguments I make in the coercion book.

The difference is, in *that* book, I have the reader's attention on the topic of coercion, and can properly address the pros/cons. In this book, they're not thinking about it, which makes confusion about such things potentially more dangerous.

So, I would judiciously use coercion in any other book context, but not just blindly because I wrote about such techniques in a different book.

I'm not going to make this change now, but I'll reconsider its' treatment for second edition. Thanks!

PR #630, opened 2016-01-10
adding previous chapter, next chapter and table of contents links for each chapter

elie:
It's slightly frustrating to jump from chapter to chapter when reading on GitHub - these links allow you to move back and forth easily from chapter to chapter.

### DIFF
--- a/async & performance/ch1.md
+++ b/async & performance/ch1.md
@@ -1,3 +1,5 @@
+#### [Table of Contents](toc.md#table-of-contents) | [Next Chapter ⇒](ch2.md)
+
 # You Don't Know JS: Async & Performance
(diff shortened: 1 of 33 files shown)


getify:
I appreciate the effort to improve readability of this content, but I don't think I want that stuff in these files. They are primarily for providing publishable content for O'Reilly and this is stuff that the editing staff would have to remove.

It's still something I'd eventually like to do to make this content more "readable" in its current online format, sorta like what gitbook does, but I haven't scoped much time for that yet. [I have an idea](http://github.com/getify/emdash) of what it would be like, though. Unfortunately gitbook isn't quite what I need.

PR #632, opened 2016-01-10
Fix typo in CH1 and CH2 from the use of an RHS/LHS to the RHS/LHS

shannabarnard:


MehdiRaash:
It's about a while I've been thinking about the indefinite articles, the used and misused of that.
Grammatically you are right (an + vowel sound), but you know it doesn't feel Okay on my tongue when I say "**a LHS**". "**an LHS**" is quit better!
Just like "a university", and as you maybe know "an university" in wrong!

shannabarnard:
@MehdiRaash Thank you for the feedback. I am familiar with what you are referring to: 

When to use 'an' or 'a' depends on how the word is pronounced and not on how it is written.


>The U in university is pronounced with a long ‘u’ sound which sounds like ‘yew’ and is written as j in the phonetic alphabet. So, although the letter is a vowel, it is not pronounced like one in 'university' because it does not have a vowel sound. We therefore say ‘a university’.
The U in umbrella is pronounced as a vowel sound ( Λ using the phonetic alphabet) and so we use ‘an’. We therefore say ‘an umbrella’.

Having looked into this further I believe 'a' is still better than 'an' but the correct grammer would be to use 'the'; the LHS and the RHS. I will make the suggested update.

MehdiRaash:
So as you said "...is depends on the words is pronounced" "a" would be the correct choice! in these cases.
And about "**the**": I think(or maybe feel), the author by saying "an LHS"(for example) was trying to announce somehow like *an instance of LHS look-up* or generally an LHS look-up, not exactly recent LHS look-up he talked about.
It's a bit tricky but I think "**the LHS**" instead of LHS with indefinite article doesn't convey that, an LHS does more sound **in general**.

getify:
I don't think this necessitates a change, but thank you for the suggestion.

PR #644, opened 2016-01-23
Feature/update table of contents with relative links

emkk:
This series has been just what I was looking for. Thanks for writing it. 

Currently digging in and sometimes feel it is a little difficult getting back to my spot at times, especially on mobile.

So I've added relative links based on GitHub markdown to all the `toc.md` files. Would a top-level `toc.md` file, essentially an amalgam of all the `toc.md` files from the various books, be useful?

Best,
Edmund


getify:
Please see: https://github.com/getify/You-Dont-Know-JS/pull/630

PR #648, opened 2016-01-29
Updated bases for number calculations

ramadis:


getify:
Hmmmm.... I'm not so sure about this.

Yes, `Number(..)` and `+` can take properly formatted non-10 base numbers (like base-2, base-8, and base-16).

But the intent of this overall section is to talk about implicit coercion, like that which would happen with, say, the `/` operator. **THAT** kind of (implicit) coercion only accepts base-10 like values.

Thinking maybe the less misleading way to "fix" this is to remove the reference to base-16 (leaving only base-10) rather than add the other bases.

getify:
[edit: inaccurate info removed]

getify:
Upon reflection, I think this needs more explanation, that the values in their other non-10 bases have to be formatted correctly (aka as the numeric literal would be -- "0b00110101010" not just "00110101010").

PR #652, opened 2016-02-01
Update ch6.md

MehdiRaash:
persistence on a good habit, using `var`.


getify:
Good point, thanks!

MehdiRaash:
My pleasure.

PR #663, opened 2016-02-14
fix grammar in "this & object prototypes" chap. 4

ncpierson:
"classes" is plural, even though it is quoted.

http://www.quickanddirtytips.com/education/grammar/oddness-when-you-start-a-sentence-with-there-is


getify:
Thanks!

PR #669, opened 2016-02-21
fixing value returned from generator

magul:
I'm not certain, that this is proper fix from Your POV but I was trying to not include any other features here and print plain object, that generator actually returned.


getify:
Thank you!

PR #670, opened 2016-02-22
Removed 'amount' modification in example to be consistent with narrative

SIRHAMY:
-Narrative is discussing the addition of a $9.99 add on to the price. There is no reason to multiply the base amount ($99.99) by 2.


getify:
That line is there on purpose to illustrate that the conditional is based on the current value of the variable as it changes over time during the program.

PR #673, opened 2016-02-24
Update ch5.md to reflect ES2015 (ES6)

George8211:
The info on strict mode throwing an error for duplicate property names in an object literal didn't mention that it's been removed in ES2015 because of computed property names, so I've added it.

Would it be good to include an example for computed property names in an object literal to explain?


getify:
Interesting, thanks for the update. I'll reserve this update for second edition.

Issue #675, opened 2016-03-01
"types & grammar": Pre-ES6 Number.isSafeInteger(..) polyfill - Number.MAX_SAFE_INTEGER is also introduced ES6


johnBartos:
In Chapter 2 of "types & grammar", the following polyfill is prescribed for `Number.isSafeInteger(..)`

> To polyfill Number.isSafeInteger(..) in pre-ES6 browsers:
> 
> ```
> if (!Number.isSafeInteger) {
>     Number.isSafeInteger = function(num) {
>         return Number.isInteger( num ) &&
>             Math.abs( num ) <= Number.MAX_SAFE_INTEGER;
>     };
> }
> ```

Unless I'm mistaken, isn't Number.MAX_SAFE_INTEGER also an ES6 feature? Perhaps the code should rather be:

```
if (!Number.isSafeInteger) {
    Number.isSafeInteger = function(num) {
        return Number.isInteger( num ) &&
            Math.abs( num ) <= (Math.pow( 2, 53 ) - 1)
    };
}
```

getify:
The polyfill is illustrative rather than prescriptive. Use the polyfills from [es6-shim](https://github.com/es-shims/es6-shim) for real production usage.

johnBartos:
Of course - my issue is that it seemed inconsistent to introduce two ES6 features and recognize only one in a polyfill which does requires both. It's really just nitpicking though :ok_hand:

getify:
In a future edition I'll try to make this a little more clear. Thanks.

PR #677, opened 2016-03-02
fix(types and gramar - chapter 2)

deezahyn:
- polyfill for `Number.isSafeInteger(..)` in pre-ES6 browsers, is using ES6 only constant: Number.MAX_SAFE_INTEGER
- this fix uses number literal instead


getify:
Thank you, but I'm going to [defer handling this until second edition](https://github.com/getify/You-Dont-Know-JS/issues/675) and do so in a bit of a different way.

deezahyn:
@getify - My bad, I missed the open issue.  Loving types and grammar so far, the whole series really.  Thanks for this awesome book!

getify:
:+1:

PR #684, opened 2016-03-11
Update ch2.md

simonebrunozzi:
Specified what's the new type introduced in ES6.


getify:
Thank you for the clarification. :)

PR #685, opened 2016-03-11
Update ch2.md

simonebrunozzi:


getify:
There's currently only one type that qualifies as a number. No need to make this change.

PR #687, opened 2016-03-15
Coerces truthy values to Bool in onlyOne()

galenweber:
The current version of the implicit coercion form of onlyOne() cannot handle truthy values. For example, onlyOne("42", false) will return false.


galenweber:
I see that this has already been addressed by getify in pull request #410 ('Fixed onlyOne() not working as intended').

I agree with getify's point there, perhaps the word "truthy" could be dropped from the sentence below?

>This `onlyOne(..)` utility should only return `true` if exactly one of the arguments is `true` / truthy.

PR #690, opened 2016-03-17
Added 'o' for octal representation

PradipShrestha:


getify:
The original text is correct. Note the "note" paragraph immediately after, explaining.

PR #695, opened 2016-03-20
Unnecessary usage of toLowerCase().

magul:
val already contains onlu lower cases because it was forced few lines above.


getify:
Thanks!

PR #696, opened 2016-03-21
There doesn't exist JSON.stringification(..).

magul:
`JSON.stringification(..)` here is formated as formal method of JSON object. It should be changed to `JSON.stringify(..)` as this method actually exists or change to informal, normal text `JSON stringification`.


getify:
Thanks!

PR #697, opened 2016-03-21
Fixing issues with custom Object.observe listing

magul:


getify:
Not sure we need to make any changes here since object.observe has been withdrawn as a proposal and will eventually be removed. In the second edition, I plan to remove this entire section on it and leave a note of deprecation.

magul:
I was unaware of that. It's basically Your call, what to do here, but untill second edition will arrive, there will be an error here (at least in comparison with how it works in Chrome and how it is define on MDN https://developer.mozilla.org/pl/docs/Web/JavaScript/Reference/Global_Objects/Object/observe).

magul:
If You will decide not to include this change, please just close PR.

PR #700, opened 2016-03-24
"async & performance" - ch 3 - typo in "Thenable Duck Typing"

johnBartos:
It looks like there was a bit of an error deciding which phrasing to use, and both ended up in the doc. I think "inability to change" sounds better.

### DIFF
--- a/async & performance/ch3.md
+++ b/async & performance/ch3.md
@@ -422,7 +422,7 @@ Both `v1` and `v2` will be assumed to be thenables. You can't control or predict
 
 Sound implausible or unlikely? Perhaps.
 
-But keep in mind that there were several well-known non-Promise libraries preexisting in the community prior to ES6 that happened to already have a method on them called `then(..)`. Some of those libraries chose to rename their own methods to avoid collision (that sucks!). Others have simply been relegated to the unfortunate status of "incompatible with Promise-based coding" in reward for their inability to change to get out of the way.
+But keep in mind that there were several well-known non-Promise libraries preexisting in the community prior to ES6 that happened to already have a method on them called `then(..)`. Some of those libraries chose to rename their own methods to avoid collision (that sucks!). Others have simply been relegated to the unfortunate status of "incompatible with Promise-based coding" in reward for their inability to change.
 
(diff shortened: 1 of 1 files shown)


getify:
The wording as-is was intentional. "to get out of the way" is reinforcing the idea that the *change* is simply to avoid the inevitable paving over that is happening with the `then` namespacing. Your change is not bad, but the original wasn't a mistake.

johnBartos:
I see - I believe it's a bit of a [garden path sentence](https://en.wikipedia.org/wiki/Garden_path_sentence) as is though. Perhaps a conjunction? 

```
their inability to change (and/in order to to) get out of the way
```

getify:
Garden Path? No, I don't think so. I don't see how it's being mis-parsed. It reads clearly to me. I think it's a colloquialism, for sure. It's definitely my informal speaking tone coming through in my writing.

Consider this sentence: "I had to move to get out of the way." Does that sentence make sense, or does it bother you in the same way? That sentence could be "I had to get out of the way", which just implies the "move" part that most readers will understand, I'd guess.

In my book's sentence, I used "change" instead of "move", because the point is a alteration in software design to avoid collisions with other software design. But similarly, I could have dropped "to change" and said just "...inability to get out of the way". But there I think the implication of "change" is far less obvious, and makes the sentence, IMO, less understandable.

johnBartos:
`"I had to move to get out of the way."`

The difference is that `move` is a verb while `inability` is a noun. This results in a fundamentally different sentence - the latter has no subject getting out of `the way`, while the former does. Consider the following from the Stanford parser:

```
to move to get out of the way

(ROOT
  (FRAG
    (S
      (VP (TO to)
        (VP (VB move)
          (S
            (VP (TO to)
              (VP (VB get)
                (PRT (RP out))
                (PP (IN of)
                  (NP (DT the) (NN way)))))))))))
```

```
I had to move to get out of the way

(ROOT
  (S
    (NP (PRP I))
    (VP (VBD had)
      (S
        (VP (TO to)
          (VP (VB move)
            (S
              (VP (TO to)
                (VP (VB get)
                  (PRT (RP out))
                  (PP (IN of)
                    (NP (DT the) (NN way))))))))))))
```

```
their ability to move to get out of the way
(ROOT
  (NP
    (NP (PRP$ their) (NN inability)
      (S
        (VP (TO to)
          (VP (VB move)
            (S
              (VP (TO to)
                (VP (VB get)
                  (PRT (RP out)))))))))
    (PP (IN of)
      (NP (DT the) (NN way)))))
```

Issue #701, opened 2016-03-24
`arguments` spoken as if it was an array


oldergod:
In _[Types & Grammar / Chapter 5: Grammar](https://github.com/getify/You-Dont-Know-JS/blob/master/types%20&%20grammar/ch5.md)_, `arguments` is refered many times as being the "`arguments` array" while it is, as I am sure you know, an object.

Would it not be better to say clearly `object` or at least `array-like`?

getify:
Shrugs. I don't think it's terribly important at those parts of the text to introduce distraction around "things that look like arrays but aren't." the "arguments array" is wording i've heard in common usage before.

PR #708, opened 2016-04-05
Replace 'arguments array' with 'arguments object' 

dharam:
I guess referencing 'arguments object' as an array adds some scope for confusion.
Kindly merge it if it seems fit.


oldergod:
fix #701

PR #710, opened 2016-04-06
Script for automatically generating epubs + support files

nphyx:
I wrote a quick bash script and some metadata files to turn these in to epubs so I could read them on my tablet. I'm not sure if you or your publisher has the epub rights, but if you are allowed to publish as epub I figured this might save you some work. You'll need pandoc installed (http://pandoc.org).

The script could probably be improved a bit to be more robust; I've only tested in Arch Linux. It wouldn't be hard to adapt it to other output formats pandoc supports.

I've added the output to .gitignore so that commits won't include the built epubs.

Additionally, the metadata.xml files may need to be tweaked to your liking.


getify:
Appreciate this, but O'Reilly holds the exclusive rights to distribute the book content as "ebook" (incl epub). The content is only distributed in rendered markdown here under CC3.

nphyx:
No problem. Would you like me to remove the fork containing it? Don't want to cause trouble with your publisher.

getify:
Nah, it's fine. Just wanted to explain why I couldn't officially merge/endorse it in the main repo. :)

nphyx:
Cool cool.

PR #716, opened 2016-04-22
Mistake on writable property

drizko:
If something is not marked as read-only then writable would presumably be true.


getify:
Dupe. https://github.com/getify/You-Dont-Know-JS/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aclosed+writable

PR #717, opened 2016-04-24
Replace: anti-climatic with anticlimactic

ascasson:
Just a simple word swap. Anticlimactic is a word meaning without a climax, whereas climatic will refer to weather/climate. 

P.S. Thanks so damn much for this book series! I love it.


getify:
Thanks!

PR #718, opened 2016-04-25
Update ch2.md

infinnie:
Italic non-keyword *this*


getify:
If I use "this" in a way that refers to the keyword, it's always marked up as `this`. Otherwise, it's just a normal grammatical pronoun. Italicizing it doesn't really add any clarity in this case. There are a few places where it's added a stylistic flourish to the text, but I don't want to go overboard.

PR #722, opened 2016-04-27
"async & performance" - ch4 - generator order fixes

davloperez:
I believe the "order" in which generator are called (from outside to inside) is: 
*bar() -> *foo(3) -> *foo(2) -> *foo(1)


getify:
Thanks, but dupe of https://github.com/getify/You-Dont-Know-JS/pull/434 :)

PR #724, opened 2016-05-04
Fix missing note

umpox:
[^note-let_er]: [let-er] will not display anything as a space is required after the text 'let-er'. I have added this space and the note now appears at the bottom of the page.

### DIFF
--- a/scope & closures/apB.md
+++ b/scope & closures/apB.md
@@ -120,4 +120,4 @@ The question really becomes: do you want block-scoping, or not. If you do, these
 
 [^note-traceur]: [Google Traceur](http://traceur-compiler.googlecode.com/git/demo/repl.html)
 
-[^note-let_er]: [let-er ](https://github.com/getify/let-er)
+[^note-let_er]\: [let-er](https://github.com/getify/let-er)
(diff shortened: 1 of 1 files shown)


getify:
why is a space required?

umpox:
@getify Markdown thinks you are trying to add a reference link and expects `[let-er]` to be a URL, it is not so it does not display. This is because the line begins with `[^note-let_er]:` which is the beginning of the syntax for a reference link. I wasn't happy with the solution to add a space after the text so I have added a backslash escape to break the syntax and ensure Markdown treats `[^note-let_er]:` as plaintext.

umpox:
Accidentally closed.

getify:
why doesn't the traceur link above it have the same problem?

umpox:
Markdown expects everything within `[ ]` to be a link, when a space is added this syntax breaks and it is treated as just text.

For example:

`[reference text]: [https://www.mozilla.org]` will not display (although elsewhere writing [reference text] will automatically hyperlink to mozilla. 

`[reference text]: [https://www.mozilla.org ]` will now display as a space has been added after the link breaking the syntax

getify:
OK, thanks!

PR #729, opened 2016-05-13
I suggest the following edit on line 455

Sefton419:
Because all other `setTimeout( )` functions in this chapter contain a space after the comma between parameters (528 - 530, 836 - 839, 857, 1107 - 1109, 1127 - 1135, 1727 - 1729), I suggest the following edit on line 455:

`setTimeout(..,0)` --> `setTimeout(.., 0)`


getify:
I don't add the extra spaces in inline code segments, only in the code blocks.

PR #731, opened 2016-05-15
Update ch5.md: small typo

dmitriz:
Only `hello` is imported, not the the result of its invocation `hello()`


getify:
I use the `()` notation throughout the text to mean that we're referring to a function, not that we're calling it.

PR #739, opened 2016-05-19
Console log minor changes

zcserei:
TBH, I don't quite understand why the quotes appear in the log (they are single quotes while running under node and double quotes in Chrome), I'd actually appreciate a comment on this one.


getify:
There's no standard that controls how `console.log` output will appear. It's different for each environment.

zcserei:
I see.

Thank you again for the clarification. I will try to RTFM before creating a request next time! 😵

getify:
Cheers. :)

PR #745, opened 2016-05-26
Fixed typo

ryannewton:


getify:
what's printed comes from the `console.log(..)` statements, not from the results of the expressions. they're deliberately reversed.

PR #752, opened 2016-06-03
Typos with effect / affecting

Isocles:
EDIT: This is in Types and Grammar Chp 5, notably this section: 
https://github.com/getify/You-Dont-Know-JS/blob/master/types%20%26%20grammar/ch5.md#expression-side-effects

I notice you refer to operators with side effects a lot, and use the verb `side-effecting` while I can understand that it would be logical to simply add `ing` to the end of `side-effect`, that verb is used differently. Technically, you should be using `affecting` instead. However, one could make the case that using the term `side-effecting` makes it clearer for the readers to understand that it is simply a reference to operators with side-effects, which are defined beforehand. It's up to you to decide, though I submitted the orthographically correct changes in this commit in case you want to correct it.

### DIFF
--- a/types & grammar/ch5.md
+++ b/types & grammar/ch5.md
@@ -139,7 +139,7 @@ var a = 1;
 foo();		// result: `undefined`, side effect: changed `a`
 ```
 
-There are other side-effecting expressions, though. For example:
+There are other side-affecting expressions, though. For example:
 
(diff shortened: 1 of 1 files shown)


Isocles:
Upon further research, effecting could actually be used.

PR #753, opened 2016-06-03
fix typo

Isocles:
added "al" to "mathematic" in the expression "Any mathematic operation".


Isocles:
Oops, didn't see someone already made a pull request for this fix, seen here: #751. You can just ignore this one then.

PR #755, opened 2016-06-05
Fixing Example for CH4 Types And Grammar - Abstract Relational Comparison

arseniew:
There are two issues with the following example:

> _The algorithm first calls ToPrimitive coercion on both values, and if the return result of either call is not a string, then both values are coerced to number values using the ToNumber operation rules, and compared numerically._
```javascript
var a = [ 42 ];
var b = [ "43" ];
 a < b; // true
 b < a; // false
```

First of all, comparison will yield exact same value when you compare numerical and string values, so:
```
"42" < "43"; // true
42 < 43; // true
```

Which masks the second problem: In this case both of`[ 42 ]` and `[ "43" ]` they will become strings after ToPrimitive.

In order to solve both of this problems this PR changes `b` value to `["143"]`, which will cause string comparison to yield different value than text comparison.
Change made to `a` value is to make sure ToPrimitive function for it will return number.

### DIFF
--- a/types & grammar/ch4.md
+++ b/types & grammar/ch4.md
@@ -1830,8 +1830,8 @@ The algorithm first calls `ToPrimitive` coercion on both values, and if the retu
 For example:
 
 ```js
-var a = [ 42 ];
-var b = [ "43" ];
+var a = { valueOf:function() {return 42;} };
+var b = [ "143" ];
 
 a < b;	// true
 b < a;	// false


arseniew:
That's not the entire change. 
Your example returns false because in that case string comparison is used, so it boils down to: `"42" < "143"`

Take a look at entire change made: https://github.com/getify/You-Dont-Know-JS/pull/755/files

getify:
I see why the example and preceding text are a bit misleading when paired together. But the fix I'd prefer is probably much simpler:

```js
var a = new Number( 42 );  // number object
var b = [ "43" ];
a < b;     // true
b < a;     // false
```

arseniew:
Agreed. Number object does way more sense in this example.
I would still prefer different value for `b` to prove that both values are converted to numbers, rather than string.

getify:
Will figure out how to address for second edition.

animeshk874:
I personally think that a better example would be:

`
var a = [42];
var b = ["143"];
var c = new Number(143);

a < b; //false because of string comparison
a < c; //true because of numeric comparison
`

PR #758, opened 2016-06-10
Update ch3.md

LukaszBrygala:
I think that is correct output of console.log(x) ?


getify:
Yep, good catch... I think this might have been an artifact from pre-implementation testing. Thanks!

getify:
Looks like it was fixed before print, but must have failed to get back-ported to this repo.

PR #764, opened 2016-06-15
Fixed description of typed array constructor

LukaszBrygala:
I think that is the proper description? Not `length` bytes, but rather `length` elements initialized to 0.


getify:
Thanks.

PR #768, opened 2016-06-21
Update ch3.md

marioasu:
there's a clerical error


getify:
why are you removing 'var' in just one place? it's intended.

marioasu:
i know 'var j' is correct.but in this place, you want to explain pick another identifier is an optional but not good solution. if you add var in front of the variable,then you dont have to change the variable i to j. 'var i' is ok, just like the previous solution(good solution). that's my apprehend,please review the whole paragraph more carefully.

getify:
The `j` would have to be declared with a `var` somewhere. If not there, where would you propose it be declared?

marioasu:
when you want to example a bad practice (just only change i to j), you dont need declared with a var. var i,var j,var something else is a good practice but has nothing to do with change variable.

getify:
If you just start assigning to an undeclared `j`, you'd either get an error in strict mode or an accidental-global in non-strict mode. Neither of those is a good idea to illustrate.

PR #769, opened 2016-06-21
Use strict-equals rather than loose-equals in example

gsklee:
Would be great to let beginners befriend `===` rather than `==` upon their first encounter with the equality operators.


getify:
i don't think that change adds anything to the text. Thank you though.

PR #776, opened 2016-06-27
variable values corrected in comments

saraf:
The variable values were incorrect in the comments in an example in the "Statement Ordering" section.


getify:
The comment output is not the value of the variables... it's the `console.log(..)` statements in the corresponding getters/setters.

saraf:
Er, I know that, but those are the other way around.

PR #778, opened 2016-06-28
Update module names for consistency: foo -> A, bar -> B

airandfingers:
Lines 1540 through 1573 refer to the module with "foo" as "A" and the module with "bar" as "B".

Lines 1580 through 1605 should match these names, rather than referring to "A" as "foo" and "B" as "bar".


getify:
Thanks. Queued for second edition updates.

PR #779, opened 2016-06-28
fix grammar error (incomplete sentence)

jonbri:
"why" needed to complete sentence


getify:
Thank you, but I'm not going to make this change.

PR #780, opened 2016-06-28
Easily movable table of contents to up&going

uppusaikiran:
When i was going to readme.md file,because it is so large i often lost where iam presently there.So entire chapter into several sub modules and these sub modules are listed as table of contents and links to appropriate sub modules are added.

### DIFF
--- a/up & going/ch1.md
+++ b/up & going/ch1.md
@@ -13,6 +13,19 @@ Once you feel comfortable with general programming basics, Chapter 2 will help g
 
 If you're already fairly comfortable with JavaScript, first check out Chapter 3 as a brief glimpse of what to expect from *YDKJS*, then jump right in!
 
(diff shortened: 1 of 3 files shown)


kevcomedia:
I'm also thinking that a small table of contents in each chapter would be useful, since a chapter can be fairly long.

kevcomedia:
Hi again. I saw your changes and noticed that you used absolute URLs in the links. In my opinion a relative URL is a better choice.

uppusaikiran:
Thanks for suggestion , iam awaiting response from owner.Because he might be in another idea.

getify:
The primary intent of this repo is to maintain the text so that it can be imported by the book publisher and produced. I do not arrange it or add content here which is not appropriate for that task because then it's a lot of manual error-prone work to filter it out whenever they import.

PR #785, opened 2016-07-03
Update link for Traceur

slivingston:
The Traceur repository is now hosted on GitHub. The original link to a site on Google Code is not valid anymore because the hosting service ended, which was announced at http://google-opensource.blogspot.com/2015/03/farewell-to-google-code.html

### DIFF
--- a/scope & closures/apB.md
+++ b/scope & closures/apB.md
@@ -118,6 +118,6 @@ Secondly, IIFE is not a fair apples-to-apples comparison with `try/catch`, becau
 
 The question really becomes: do you want block-scoping, or not. If you do, these tools provide you that option. If not, keep using `var` and go on about your coding!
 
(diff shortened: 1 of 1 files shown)


getify:
Thanks.

brianbishop:
In Appendix B: Polyfilling Block Scope, where that link is used, the [Traceur online compiler](https://google.github.io/traceur-compiler/demo/repl.html) does not traspile to a `try{}catch{}`, and uses identifier renaming instead: 

```javascript
var a = 1;
{
  let a = 2;
  console.log(a);
}
console.log( a );
```

```javascript
var a = 1;
{
    var a$__0 = 2;
    console.log(a$__0);
}
console.log(a);
```

legodude17:
@brianbishop you should probably file a separate issue.

getify:
Closing, as the 2nd-ed updates will not mention Traceur.

PR #787, opened 2016-07-03
Placed links at the bottom of each chapter to the next

RussellAFisher:
Placed links in each chapter to the next chapter or appendices. All locally hosted forwards link to the first chapter of their book. Preface is stand alone so no link on that. Last chapter of each book links all appendices instead of chaining them.


getify:
The primary intent of this repo is to maintain the text so that it can be imported by the book publisher and produced. I do not arrange it or add content here which is not appropriate for that task because then it's a lot of manual error-prone work to filter it out whenever they import.

PR #795, opened 2016-07-12
Add missing word

regularmike:
This PR adds a missing "how" to the description of the iteration helpers' behavior. 

"Each of these helpers accepts a function callback to apply to each element in the array, differing only in *how* they respectively respond to a return value from the callback."


getify:
Thanks

PR #797, opened 2016-07-15
Add a return statement

ivandotv:
Add `return` statement in order to be consistent with the previous example for `function bar()`


getify:
Good catch! Thanks!

PR #802, opened 2016-07-19
Update preface.md

DanAtShenTech:
Fix typo - "now" should be "know".  Add italics as appropriate.


getify:
"now" is the intended word. I'm saying that these books *now* help you learn the language. If it says "know", then that word is repeated twice in the sentence, which I think would be more confusing on the whole.

PR #804, opened 2016-07-21
Update ch2.md

Tiramisusie:
fix a formatting problem


getify:
The current formatting is intended.

PR #805, opened 2016-07-21
Update ch5.md

kevcomedia:
I think the parenthesized code should pertain to the emphasized words before it.

It's confusing to see (`writable:false`) before the phrases "and it's not marked as read-only" and "but it's marked as read-only".


getify:
Known issue: https://github.com/getify/You-Dont-Know-JS/issues?utf8=%E2%9C%93&q=is%3Aissue%20writable

PR #807, opened 2016-07-23
Changing "module" keyword to "import"

yammesicka:
After trying and searching in multiple resources, it seems to me that there is no "module" keyword in JavaScript.
The description in the book, which claims that:

> `module` imports an entire module API to a bound variable

Is true also for the [import statement](https://developer.mozilla.org/en/docs/web/javascript/reference/statements/import), so I've changed it accordingly.


getify:
This already a known issue (spec changed after publication) and will be addressed in second edition. Thanks.

PR #810, opened 2016-07-25
es6+beyond: ch2.md, fixing word order issue

gryffs:
Small word order change.  Original text:
"is... very not ideal, performance wise."
screen shot 2016-07-24 at 8 33 52 pm

Changed to read:
"is... not very ideal, performance wise."
screen shot 2016-07-24 at 8 51 23 pm

### DIFF
--- a/es6 & beyond/ch2.md
+++ b/es6 & beyond/ch2.md
@@ -2533,7 +2533,7 @@ var s1 = "abc\u0301d",
 [...s3.normalize()][2];			// "𝒞"
 ```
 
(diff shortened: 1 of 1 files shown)


gryffs:
This may also read better as "... far from ideal" or just "... not ideal," but the change above was submitted with the assumption that the original wording was simply ordered incorrectly.  Hope this helps.

getify:
Thanks. :)

Numpung25:
hello

PR #812, opened 2016-07-25
es6+beyond: ch3.md, fixing typo

gryffs:
Update to the wording in es6 & beyond: ch3.  If this was your intended meaning, I hope it helps :)
Previous wording:
"as any `yield ..` expressions in the `finally` clause are respected to pause and send messages"
screen shot 2016-07-25 at 8 23 17 am

Intended wording:
"as any `yield ..` expressions in the `finally` clause are expected to pause and send messages"

screen shot 2016-07-25 at 8 32 24 am

### DIFF
--- a/es6 & beyond/ch3.md
+++ b/es6 & beyond/ch3.md
@@ -843,7 +843,7 @@ it.return( 42 );		// cleanup!
 						// { value: 42, done: true }
 ```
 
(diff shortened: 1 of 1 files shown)


getify:
The original wording is intended. I'm saying that any `yield`s inside a `finally` are still handled (aka "respected") even though it might have seemed like they would not.

gryffs:
OK, I see it.  Thank you for the clarification and your time!

PR #814, opened 2016-07-26
Error in example.md

robisrob:
On of the two values must have not a 'string' as the result of the 'ToPrimitive' coersion. 

When ToPrimitive is called on an array, the result will always be a string, because the valueOf is returning the array itself. When this method isn't returning a primitive value, it will try using toString().


getify:
Known issue: https://github.com/getify/You-Dont-Know-JS/pull/755

PR #829, opened 2016-08-02
corrected error in hoisting example

Huanzhang89:
if the variable declaration is hoisted first, the function declaration is treated as a duplicate and thus ignored. At the point of invoking `foo()`, the value of `foo` is `undefined` and therefore should throw a `TypeError`.

### DIFF
--- a/scope & closures/ch4.md
+++ b/scope & closures/ch4.md
@@ -158,7 +158,7 @@ foo = function() {
 };
 ```
 
-`1` is printed instead of `2`! This snippet is interpreted by the *Engine* as:
+`1` is printed instead of a `TypeError` from trying to invoke `undefined`, which is not a function! This snippet is interpreted by the *Engine* as:
 
 ```js
 function foo() {


getify:
This is incorrect. Functions "hoist" first and *can* override duplicates. Then vars "hoist" and cannot. override.

Huanzhang89:
Hi Kyle,

I re-read my PR and have realised that it may be a bit unclear on what I was trying to say. I haven't made many contributions to public repositories so far so I am still learning the etiquette etc. I have definitely learnt a lot from your books and did not mean to sound like I was challenging you. 

I actually made this PR because I found one of the examples on hoisting from the Scope and Closures book a little confusing from the point of view of a junior developer, since I expected that foo() will actually error if variable declarations are hoisted first. In the case that function declarations can override duplicates, then I would expect the console log resulting from `foo()` to still be 1. 

With that in mind may I offer an alternative example that I think would be easier to understand.

```
var foo;

foo = function() {
    console.log( 2 );
};

foo(); // 1

function foo() {
    console.log( 1 );
}
```

PR #832, opened 2016-08-05
clarify which snippet is which

kaycebasques:
I had to jump between the code and the text a couple of times to figure out which snippets you were talking about.


kaycebasques:
I think, at least! I'm not even sure if I have this correct.

getify:
Thanks for the suggestion... will address for second edition.

PR #834, opened 2016-08-10
An uncaught TypeError will be thrown

SDSethia:
An uncaught TypeError will be thrown if this is ran on a fresh console. Any in any case "b" will never be printed.


getify:
ref: https://tc39.github.io/ecma262/#sec-block-level-function-declarations-web-legacy-compatibility-semantics

getify:
also: https://twitter.com/awbjs/status/773578050654965760

getify:
Closing as not applicable anymore, since the scope & closures 2nd-ed has been rewritten.

PR #837, opened 2016-08-14
Ch4:Change log statement and remove //yep comment.

peterheesterman:
As it currently reads it seems to indicate to the reader that the if block will execute. It wouldn't.

As i read it, a && d is 42 && null which results in null.
then d || c is null || undefined which results in undefined.

undefined is falsy so the console statement will not execute.


getify:
please look at the value of `c` that is set 4 lines up.

peterheesterman:
Thanks, sorry for inconvenience.

getify:
No worries.

PR #838, opened 2016-08-16
Make a letter uppercase to match list style

TrevorEdwards:
"swallow" had inconsistent casing from its list, so this fixes that error.


getify:
Thanks!

Issue #816, opened 2016-07-27
Fibonacci iterator example


mahmoudfelfel:
Hi Kyle,
Thanks for the amazing book series.
I'm reading now the "ES6 and Beyond" part, it's amazingly detailed, but I have some questions about the Fibonacci iterator example in ch3:

```
var Fib = {
    [Symbol.iterator]() {
        var n1 = 1, n2 = 1;

        return {
            // make the iterator an iterable
            [Symbol.iterator]() { return this; },

            next() {
                var current = n2;
                n2 = n1;
                n1 = n1 + current;
                return { value: current, done: false };
            },

            return(v) {
                console.log(
                    "Fibonacci sequence abandoned."
                );
                return { value: v, done: true };
            }
        };
    }
};

for (var v of Fib) {
    console.log( v );

    if (v > 50) break;
}
```

Why we need another ```[Symbol.iterator]() ``` inside the first one, we just need ```next()``` to implement an iterator, and the wrapper ```[Symbol.iterator]() ``` is to create an iterable of our object so we can use it with ``` for .. of```.

Another question, isn't it better to stop the loop implicitly inside the implementation instead of ``` if(v>50) break; ```, we can do :

```
           next() {
                var current = n2;
                n2 = n1;
                n1 = n1 + current;
                console.log(current)
                if(current > 50) return this.return(current)
                return { value: current, done: false };
            },
```
So we can avoid the infinity loop, i mean in real life coding, if someone reuses this function, he might forget to add the ```break``` line which will break the code.

Thanks

legodude17:
The `[Symbol.iterator](){ return this; }` is so that the iterator is also iterable. The reason he did `if(v<50)break;` in it is so the iterator can be used for different amounts.

getify:
Yes, that's correct @legodude17. So that you can do:

```js
var it = Fib();

for (var v of it) {
   console.log(v);
   if (v > 50) break;
}
```

The iterator doesn't have its own limit baked into it, as that'd be entirely inappropriate. The consumer (the `for..of` loop) is the place someone should decide how many numbers they want.

mahmoudfelfel:
Thanks @legodude17 @getify for your answers, but ```for .. of``` requires only ```Fib()``` to be an iterable, and we accomplished that with the first ```[Symbol.iterator](){ return .. } ``` so why we need the returned iterator to be an iterable as well, or it is just a good practice to make the iterator an iterable.
This is giving the same result :
```
var Fib = {
[...]

PR #842, opened 2016-08-22
No need for  [Symbol.iterator]() { return this; }, here ?

UnbearableBear:
I don't understand why we should make the the Fib iterator an iterable here since Fib is already an iterable ? It's more confusing than anything IMO.

### DIFF
--- a/es6 & beyond/ch3.md
+++ b/es6 & beyond/ch3.md
@@ -194,9 +194,6 @@ var Fib = {
 		var n1 = 1, n2 = 1;
 
 		return {
-			// make the iterator an iterable
-			[Symbol.iterator]() { return this; },
-
 			next() {
 				var current = n2;
 				n2 = n1;


getify:
Already explained [here](https://github.com/getify/You-Dont-Know-JS/issues/816).

UnbearableBear:
Cristal clear. Thanks for the quick answer. Since it's the second time the question is asked, might add the answer right into the book ?

getify:
It's covered in the text in that same chapter:

https://github.com/getify/You-Dont-Know-JS/blob/master/es6%20%26%20beyond/ch3.md#iterator-loop

UnbearableBear:
As stated [here](https://github.com/getify/You-Dont-Know-JS/issues/816), the reason why `[Symbol.iterator]() { return this; }` stands here is because we might want to make a combined use of the Fib iterator: Both manually with `.next()` (with some arguments if desired) and then we might wish keep 'automatically' iterating on it with` for .. of`, until a certain value is reached.
The purpose of `[Symbol.iterator]() { return this; }` is indeed well explained in the book, but I think that the purpose of the above design pattern deserve this explanation in the book. 
Anyway, thanks for this amazing book, that I recommend to any JS dev anytime I can ^^

getify:
it's not just mixed usage that motivates the pattern. If the iterator function took arguments to use to customize its iteration behavior, you'd want to be able to manually call that function to provide arguments. But once you have the manually constructed iterator, you want it to be an iterable itself so it can be directly iterated.

getify:
Another way of explaining why it's included in the snippet in question even though not strictly needed there is to reinforce (from the section before!) the "best practice" of always making an iterator be an iterable too. There's no downside or reason you wouldn't want that

UnbearableBear:
I guess my problem here is that I needed to understand the purpose of each piece of code in the snippet to make sure I did not miss anything. And when I realized that `Fib` did not required its iterator to be an iterable to work with `for .. of`, I started having an headache ^^

PR #847, opened 2016-09-02
change book number from third to fourth

MJGwater:
My first pull request!


getify:
Thanks, and congrats on the first PR.

However, I don't think we should make this change. The Foreword was written by a contributing author, and at the time it was written, "Types & Grammar" was indeed the third book written. Though now "Up & Going" is suggested to be the "first" book to read, it was actually the fifth book published of the six.

Technically, "third installment" was not only true at the time of writing, it remains a historical fact even now -- if we take "installment" to mean "when a book published" and not "which book to read".

MJGwater:
I appreciate the thorough feedback. Really enjoying your books so far!

getify:
Cheers! Thanks again for the PR attempt.

PR #848, opened 2016-09-02
removes broken image URL

benschac:


getify:
Fixed image link already.

benschac:
Thanks, I'm really enjoying the series and the way you explain concepts! 😃

PR #853, opened 2016-09-06
foo() returns "a"

m-kermani:
**I'm not sure about this,** but we have this code:
```
foo(); // "b"

var a = true;
if (a) {
   function foo() { console.log( "a" ); }
}
else {
   function foo() { console.log( "b" ); }
}
```

and `foo()` returns` "a"`, but in the comment after` foo()` we have:

`foo(); // "b"`

But the function `foo()` returns `"a"`


viniciuskneves:
In which conditions have you tested that? Because I've just tested it and it returns "b" as expected (node v5.5.0 btw)

m-kermani:
You are right, now I tested that again and returned `b`!

I wonder why first time, I got a! I think I had changed the code and I forgot something...

PR #854, opened 2016-09-07
Keep up to date with ES spec "types & grammar" ch1.mod

dkatavic:
The ECMAScript language types quotation is missing Symbol Type. I have updated quotation from Ecma-262 Edition 5.1 to 7.0 and updated the link


getify:
This change is already queued up for second edition. Thanks.

PR #865, opened 2016-09-18
Error : Up & Going(Ch2) - Removes the var keyword

rajabishek:
With the `var` keyword while calling the `foo` function output is `undefined`. Only when the `var` keyword is not used `bar` will be declared as a property of the global object thus giving `global` as the output when the function `foo` is called.

### DIFF
--- a/up & going/ch2.md
+++ b/up & going/ch2.md
@@ -780,7 +780,7 @@ function foo() {
 	console.log( this.bar );
 }
 
-var bar = "global";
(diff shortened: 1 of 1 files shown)


getify:
That is incorrect. `var` in global scope always creates a global variable (and global object property). Be careful checking such things in non-standard environments like devtools consoles, node, or jsbin. None of those will give you accurate results.

PR #871, opened 2016-09-28
Added relative links as footer between chapters/books

holmberd:
I made this for myself to make it easier to move through the chapters and books as I was reading it. Could add "Previous Chapter" too if it is of any interest.

Thank you for a good series.


Ryalian:
In chrome, Octotree is the best extension for Github article reading

Bests,

Qijun Chen

holmberd:
@Ryalian Cheers

getify:
Thanks, but I'm not adding such things to this repo because it clutters up what's sent to O'Reilly for publication.

PR #873, opened 2016-10-01
Corrects typo

galenweber:
Changes String#normalize(..) to String.normalize(..)


getify:
This is not a typo... `String#normalize` is common notational shorthand for `String.prototype.normalize`, which is where this utility is located.

galenweber:
you're right, thanks and apologies.

PR #885, opened 2016-10-21
ch2#Non-JavaScript: elaborating on HTML5 APIs

eliranmal:


getify:
I'm not sure I want to go into this level of detail in this book. But I'll reconsider when making the second edition updates.

eliranmal:
@getify - alright, thanx for looking at it.

markxploder:
thx for that great book series )

2016-10-25 10:13 GMT+00:00 Eliran Malka <notifications@github.com>:

PR #887, opened 2016-10-22
Added links to TOC

jotasprout:
I’d have added links to what I think are anchors but don’t know how
they work in md files.


getify:
Thanks but I'm not adding such things to this repo, as they clutter up the content that gets fed to O'Reilly for publication.

PR #889, opened 2016-10-23
fix type: read instead of tread

sejoker:
Simple typo fix


deltaidea:
This is not a typo.

> **tread carefully** also tread warily
> to avoid saying or doing anything that could cause difficulties
> — [thefreedictionary.com](http://idioms.thefreedictionary.com/tread+carefully)

sejoker:
Ohh, I see.

PR #892, opened 2016-10-28
Wrong Result

muhajirframe:
The boolean coercion of previous code `((a && d) || c)` is `false` , the the statement.  I  added `!` it will be `true` and the statement will run correctly.


viniciuskneves:
It works as expected. It will return ``true`` because it does previously ``c = d ? a : b;`` so ``c`` won't be ``undefined``. Don't need this PR.

muhajirframe:
Sorry, my bad.

PR #894, opened 2016-11-04
fix ASQ.iterable() example in apB

aurelienbottazini:
I believe there is a small mistake/typo in this ASQ.iterable() example from _async & performance > appendix B_


getify:
Good catch, thanks!

PR #899, opened 2016-11-16
Difference in Node environment

Samueljoli:
Not sure if the implicit understanding is that we should follow along using the js environment on our browsers, but if not it would be good to note that console.log ( this.a ) would return undefined in the node environment. May not be obvious to newcomers and who knows maybe even 'seasoned' js devs ( Kanye Shrug ).

### DIFF
--- a/this & object prototypes/ch2.md
+++ b/this & object prototypes/ch2.md
@@ -64,6 +64,18 @@ function foo() {
 var a = 2;
 
 foo(); // 2
+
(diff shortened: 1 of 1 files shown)


legodude17:
Good catch! +1

getify:
I've made the decision not to handle node's oddities in this book series. I think it's their responsibility to either adhere more closely to expectations or at least document their own oddities better. The danger in baking this in is, a) it could change b) it sets precedent for all future aberrant envs that I have to keep track of.

PR #903, opened 2016-11-20
Typo in ch2

junyuanxue:
function should be `doLogin`.


getify:
The internal name is `doLogin()`, but the name exposed on the API is `login()`.

PR #904, opened 2016-11-20
Fix mispositioned comma.

krozycki:


getify:
This is standard style from the editor. I don't personally love it, but it's on purpose. :)

PR #905, opened 2016-11-21
Fix hyperlink to referred article

hoodwink73:
@swannodette's website returns "page does not exist" if `/` appended at the end of the url


getify:
Thanks!

PR #907, opened 2016-11-22
Ch2: Object Property Assignment Pattern mistake

filipsuk:
> But is `{ x, .. }` leaving off the `x: ` part or leaving off the `: x` part? We're actually leaving off the `x: `

In my opinion we are leaving off the `: x` part.
Am I right?

```
function bar() {
    return {
        x: 4,
        y: 5,
        z: 6
    };
}

var { bam, baz, bap } = bar();

console.log( bam, baz, bap );        // undefined undefined undefined
```

### DIFF
--- a/es6 & beyond/ch2.md
+++ b/es6 & beyond/ch2.md
@@ -595,7 +595,7 @@ console.log( x, y, z );				// 4 5 6
 
 Pretty cool, right?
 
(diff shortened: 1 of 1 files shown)


getify:
I think your snippet reinforces that you left off the `x:` part, since it's literally the `x:`, `y:`, and `z:` that are "missing" here and making the snippet return undefined.

filipsuk:
@getify You are right. It seemed to me we are leaving out the target (`: x`), because `x` in `var { x , y , z } = bar();` looked like defining the source names. Nevermind, sorry for the confusion and thanks for the book!

PR #909, opened 2016-11-23
`typeof` returns one of 6 (not 7) string values

sfletche:
"string", "number", "boolean", "undefined", "object", and "symbol"


the-simian:
function

getify:
There are 7.

venelinpetrov:
@sfletche : https://tc39.github.io/ecma262/#sec-typeof-operator-runtime-semantics-evaluation

PR #911, opened 2016-11-28
Update ch1.md

MehdiRaash:
More cleaner.

### DIFF
--- a/async & performance/ch1.md
+++ b/async & performance/ch1.md
@@ -622,14 +622,14 @@ So, we can coordinate the interaction with a simple latch, to let only the first
 var a;
 
 function foo(x) {
-	if (a == undefined) {
+	if (a === undefined) {
 		a = x * 2;
 		baz();
 	}
 }
 
 function bar(x) {
-	if (a == undefined) {
+	if (a === undefined) {
 		a = x / 2;
 		baz();
 	}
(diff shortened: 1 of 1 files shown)


getify:
I like `==`, especially since in this case it would allow either `undefined` or `null` depending on preference. I don't see this change improving anything.

MehdiRaash:
I just changed the form because:
In one of your books you gave a very nice and deep explanation to users about the **types**, I merely feel that it repeats (what you taught before and emphasise on them) that **undefined** is a type here implicitly not a string.
Thank you for the consideration.

getify:
The context of the example is that `a` is either `undefined` or it gets set to some number. No number is ever `==` (coercively equal) to `undefined`, so the `==` isn't any danger here. I tend to prefer the shorter form when it's safe, and it seems safe here.

Moreover, if someone later decided they didn't want just `var a;` but rather `var a = null;`, then the `== undefined` check actually has the benefit of `null` and `undefined` being coercively equal. So, in a way, it sorta future proofs to that case.

MehdiRaash:
Okay, "future proofs" as you said...
Mr Simpson, I know it's not a good place to say but It's about a while I have asked this question but no answer I got till now.
Would you please take a look [at this](http://stackoverflow.com/questions/40073278/google-chromes-unpredictable-behaviour-on-alert-function) question. I'm confused.

getify:
Posted a comment there. :)

PR #914, opened 2016-12-06
Typo

samueljoli:
**In fact, relative polymorphism lets us reference the base behavior from the overridden behavior.** ( In this sentence, aren't the `base behavior` and the `overridden behavior` the same? What I think it should be: `In fact, relative polymorphism lets us reference the base behavior from the overriding behavior.` )


getify:
Good catch! Thanks.

shivaylamba (commented):
Please fix merge conflict

PR #920, opened 2016-12-19
Added clarifing details for call-site in regards to Its Scope example…

jbosman:
… from chapter one.

### DIFF
--- a/this & object prototypes/ch2.md
+++ b/this & object prototypes/ch2.md
@@ -64,6 +64,12 @@ function foo() {
 var a = 2;
 
 foo(); // 2
+
+function bar() {
+	foo();
+}
+
+bar(); // 2
 ```
 
 The first thing to note, if you were not already aware, is that variables declared in the global scope, as `var a = 2` is, are synonymous with global-object properties of the same name. They're not copies of each other, they *are* each other. Think of it as two sides of the same coin.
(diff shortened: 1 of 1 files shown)


getify:
Both the `foo()` call-site and the `bar()` call-site you added have the same `this` behavior. I'm not sure I understand why this clarifies the text?

jbosman:
Hi Kyle,
First of all. Thank you so much for making this freely available. I'm loving the series and am learning so much. I've submitted this pull request in hopes that it will help out futures readers like myself. After reading the section and playing around with 'this' I still misinterpreted the default binding. I think adding what I've proposed will clear things up for people like myself. After re-reading the section I see that you don't actually say anything incorrectly. It was just the Call-Site section that starts off the chapter that led me astray. That section so greatly emphasizes to find the the call-site of the function that it let me to believe, as well as others as I've seen on a few blogs, that the call-site is what is used as the context for this. And I know you point this out as being incorrect in Chapter 1, but I think having quick example here will really hit the point home. Hope you'll take this pull-request into considering and thanks again!
-Joe

getify:
Thanks for the details.

Sounds like the confusion is about location of call-site vs the manner of the call-site. The location is entirely irrelevant, it's only the manner of the call-site (how the call is made).

Could we clarify it in that respect here and accomplish the same clarity as you're suggesting?

jbosman:
Yeah...  I was just re-reading the section and actually think you have it very well written the way it is. I think the confusion was all on my part. I just needed to read it more carefully. Sorry for the pull request. Glad to know you respond to these so quickly though! Happy holidays!

getify:
Cheers!

ithacasnowman:
I'm confused about call-site. The book defines it as "the location in code where a function is called (not where it's declared)." And it makes sense that a "site" would refer to a location, i.e., a particular line and column number in code.

But in section [Indirection](https://github.com/getify/You-Dont-Know-JS/blob/master/this%20%26%20object%20prototypes/ch2.md#indirection), you refer to `foo(), p.foo(), o.foo()` as call-sites. 

And in your comment above you say that "Sounds like the confusion is about location of call-site vs the manner of the call-site. The location is entirely irrelevant, it's only the manner of the call-site (how the call is made)." 

The more times I re-read the chapter and your comment above the more I think I understand what you want to mean, but IMO call-site is a bit confusing as it's defined and used now.

PR #927, opened 2016-12-30
Correct command output

marincelo:
The code outputs "a" instead of "b" when executed.


scope & closures/ch4.md
```diff
@@ -197,7 +197,7 @@ While this all may sound like nothing more than interesting academic trivia, it
 Function declarations that appear inside of normal blocks typically hoist to the enclosing scope, rather than being conditional as this code implies:
 
 ```js
-foo(); // "b"
+foo(); // "a"
```
legodude17: If you are going to change this, then you need to edit the paragraph above it and maybe below it.

PR #929, opened 2017-01-02
Add cross-reference to Array.from

stropitek:
In Chapter 3 of Grammar and Types, I've added a note about using `Array.from` for creating Array without empty slots, with a reference to the Chapter that goes deeper into it. I thought it would be useful since `Array.from` is both syntactically simpler and more modern way to do it.


getify:
There is no chapter 6 in this book. The chapter 6 you're apparently referring to is in a different book, the "ES6 & Beyond" book. At a minimum we'd have to make that a forward reference.

But it's a rather large jump forward in the series, in terms of expected order of reading. Generally, I try to stick to back references and not forward references.

stropitek:
Ok, my bad. Feel free to close this. Or let me know and I will change it to make it a forward reference.

getify:
I think for now I will not make the change, but I'll mark this as second edition to at least consider.

PR #933, opened 2017-01-08
md to html converted

dayitv89:
for offline use I converted *.md to *.html


getify:
That's fine but I don't want HTML in the repo.

PR #935, opened 2017-01-10
Change ES7 to EcmaScript

kentcdodds:
ES7 isn't a thing, but it normally referenced what comes after ES6 which is ES2016 which has already shipped and ES2017 will ship soon too (without `do`). So I think `EcmaScript` makes the most sense here.

I realize that this may be considered a typo. Feel free to close if you don't care :)


getify:
Not a typo, a substantial and important change. However, these sorts of updates are already queued up for the second edition refresh, whenever that happens. :) Thanks!

PR #940, opened 2017-01-12
Added buttons, easier to go to pages 

lordKnighton:

`up & going`

I've read your C Guidelines, but wanted to see if you would like this change, it's not crosslinks. If not then delete this PR with no response, perfectly fine. If you fancy it a bit, then accept it & i am willing to do the same to you other gitBooks


getify:
I appreciate the thought, but any content I include in the markdown that's not intended to be added to the print edition is one additional step of double-checking and managing the export/import with the publisher.

lordKnighton:
fair enough, doesn't hurt to try, until it does lol

PR #944, opened 2017-01-13
wrong example: heartbeat is not driven by brain

madconst:
Heart beating is not an automatic brain function. There is a phenomenon called "cardiac muscle automaticity". Pacemaker potentials are fired by sinoatrial node located in heart:
https://en.wikipedia.org/wiki/Cardiac_action_potential#Automaticity


getify:
shrugs.

PR #952, opened 2017-01-28
Update ch1.md

Dyasik:
Second part of the snippet (which is changed in this commit) is supposed to do just the same as the first part (`if (DEBUG) ...`), but checking the existence of the `DEBUG` variable. But current condition makes it log "Debugging is starting" even if the `DEBUG` variable is set to `false` or another 'falsy' value. So before logging the debugging message we need to check whether the debug is active.

### DIFF
--- a/types & grammar/ch1.md
+++ b/types & grammar/ch1.md
@@ -203,7 +203,7 @@ if (DEBUG) {
 }
 
 // this is a safe existence check
-if (typeof DEBUG !== "undefined") {
+if (typeof DEBUG !== "undefined" && DEBUG) {
 	console.log( "Debugging is starting" );
 }
 ```


getify:
Why would you define 'DEBUG = false' instead of just not defining 'DEBUG' at all?

Dyasik:
@getify I would define `DEBUG` and set it to `true` when I need debugging and to `false` if I don't need it. I thought it was the case presented.

getify:
Interesting. I wouldn't want (for example) to ship production code with a `DEBUG = false` declaration in it, as that would seem more confusing. Typically I have only `DEBUG = true` in dev environments and the absence of any `DEBUG` at all in my production env.

pahbloo:
@getify Even under these conditions, wouldn't be safer to use `if (typeof DEBUG !== "undefined" && DEBUG)` to prevent unexpected errors, especially in team projects? Murphy's Law, you know.

PR #957, opened 2017-02-10
Update ch2.md

jechante:


getify:
That `a` is there on purpose... to point out that accessing `a` and `this.a` are not the same thing and there's no crossover between the lexical and `this`.

PR #968, opened 2017-02-22
Update ch3.md

infinnie:
It should be a ReferenceError.
\_(:3」∠)\_


getify:
Good *catch* (pun intended)! Thanks.

PR #969, opened 2017-02-23
Fix typo. types & grammar/ch3.md

domingohui:


jonjensen:
I don't think this is a typo. "It doesn't help matters" is an English idiom.

getify:
Yeah, this was intentional wording. Thanks for the contribution attempt though!

PR #992, opened 2017-03-22
typo: mistakingly -> mistakenly

richardattfield:


getify:
Thanks. This was corrected in the published edition but hadn't been applied to the repo.

BTW, it appears "mistakingly" is not actually a typo but an accepted (if archaic) alternate spelling:

https://en.oxforddictionaries.com/definition/mistakingly

Anyway, best to get the repo to agree with the published edition.

PR #993, opened 2017-03-22
Fix broken links in Titles section of README.

P1xt:
Recently the links on the README to the various titles within the repository became broken. This commit replaces spaces in the subdirectory names with "\%20" which causes Github markdown to correctly display the link and for the link to properly direct to the appropriate subdirectory. (The links were working fine as of a month ago when last I checked, I assume that Github recently changed how they process links in markdown and what once worked, ceased to).

**Yes, I promise I've read the [Contributions Guidelines](https://github.com/getify/You-Dont-Know-JS/blob/master/CONTRIBUTING.md)** (please feel free to remove this line). I really did read the contributing guidelines. There didn't seem to be an issue for this and it seemed more a functional tweak than a typo that could be safely ignored until version 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,12 +15,12 @@ Please feel free to contribute to the quality of this content by submitting PR's
 
 ## Titles
 
-* Read online (free!): ["Up & Going"](up & going/README.md#you-dont-know-js-up--going), Published: [Buy Now](http://shop.oreilly.com/product/0636920039303.do) in print, but the ebook format is free!
-* Read online (free!): ["Scope & Closures"](scope & closures/README.md#you-dont-know-js-scope--closures), Published: [Buy Now](http://shop.oreilly.com/product/0636920026327.do)
-* Read online (free!): ["this & Object Prototypes"](this & object prototypes/README.md#you-dont-know-js-this--object-prototypes), Published: [Buy Now](http://shop.oreilly.com/product/0636920033738.do)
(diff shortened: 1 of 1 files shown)


P1xt:
Fixes the broken links from the main README as illustrated here:

ydkjs_titles_broken_links

getify:
I believe the change may have been from this:

https://githubengineering.com/a-formal-spec-for-github-markdown/

I don't mind updating the links per se, but I'd rather investigate first if in fact this was the culprit, and why the change, and whether this could be "fixed" by them updating the spec/parser back to the old behavior (my preference). IOW, I don't really *want* to have to %20 all my spaces in my links, that feels very un-markdown-like.

P1xt:
Thanks for linking that - and I totally understand about not wanting to %20 all your links, it does feel "messy".

Reading through the spec, it actually seems like not allowing unescaped spaces is intentional:

http://spec.commonmark.org/0.26/#link-destination

> a sequence of zero or more characters between an opening < and a closing > that _contains no spaces_, line breaks, or unescaped < or > characters, or
> 
> a nonempty sequence of characters that _does not include ASCII space_ or control characters, and includes parentheses only if (a) they are backslash-escaped or (b) they are part of a balanced pair of unescaped parentheses that is not itself inside a balanced pair of unescaped parentheses.

PR #1002, opened 2017-04-05
Async & performance Ch1 Typo

sarthak-saxena:
Fix typo reason-able to reasonable


getify:
This is not a typo... it's intentional, to emphasize something that is able to be reasoned about.

sarthak-saxena:
cool got it thanks (y)

PR #1010, opened 2017-04-14
Added new truthy example

renanmontebelo:

### DIFF
--- a/up & going/ch2.md
+++ b/up & going/ch2.md
@@ -226,11 +226,12 @@ The specific list of "falsy" values in JavaScript is as follows:
 * `null`, `undefined`
 * `false`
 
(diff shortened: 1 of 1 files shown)


getify:
Not sure why this change improves much. "this 'falsy' list" seems to clearly refer to the "list above", no?

Also, `"false"` is in fact truthy, but I'm not sure that's an example people ever get tripped up on. I've never run across it.

renanmontebelo:
The use of falsy and truthy in the same sentence made me re-read the sentence a couple of times. But I agree that does not improve much (btw Portuguese is my first language so it may be only me).

The `"false"` value is mere a heads-up that appearances can be tricky sometimes.

Thank you for your considerations.

getify:
Will leave open for future consideration. Thanks.

PR #1018, opened 2017-04-27
Clarification on the const definition

asimonia:
ES6 const does not indicate that a value is ‘constant’ or immutable. A const value can change. Instead, it is the binding that is fixed, or constant.

The following is perfectly valid ES6 code that does not throw an exception:

const foo = {};
foo.bar = 42;
console.log(foo.bar);
// → 42

### DIFF
--- a/scope & closures/ch3.md
+++ b/scope & closures/ch3.md
@@ -579,7 +579,7 @@ See Appendix B for an alternate (more explicit) style of block-scoping which may
 
 ### `const`
 
-In addition to `let`, ES6 introduces `const`, which also creates a block-scoped variable, but whose value is fixed (constant). Any attempt to change that value at a later time results in an error.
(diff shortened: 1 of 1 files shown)


eduard-sukharev:
You should update the example that follows the edited abstract, too.

asimonia:
I included an example that follows the edited abstract.

getify:
The original wording is not wrong, but it's evidence of the confusion around `const`, as "change" can either mean "reassign" (my intent) or "mutate" (which is different).

agustinaliagac:
Hi! I think that, even though the example is clear enough, the phrase "Any attempt to change that value at a later time results in an error." can potentially be misunderstood with "immutability". It's a small detail but I would still change it.

Thanks for creating this awesome repo!

PR #1020, opened 2017-04-29
Update short circuit && example to be more practical

### LINKED ISSUE #948: You Don't Know JS: Types & Grammar - "What about &&" Code example not clear enough
In **What about &&?** section we have the following code.

```
function foo() {
    console.log( a );
}

var a = 42;

a && foo(); // 42
```

result in comment should indicate that foo was called. but in fact we would get same result of "42" if first operand had been called (like in `||` operator). i think results of `a` and `foo()` should be different to make the example be clearer.

for example:

```
function foo() {
    console.log("foo");
}

var a = 42;

a && foo(); // "foo"
```

ceckenrode:
Fixes [#948](https://github.com/getify/You-Dont-Know-JS/issues/948)

### DIFF
--- a/types & grammar/ch4.md
+++ b/types & grammar/ch4.md
@@ -1234,15 +1234,15 @@ There's another idiom that is quite a bit less commonly authored manually, but w
 
 ```js
 function foo() {
-	console.log( a );
+	console.log( a * 2 );
 }
 
 var a = 42;
 
-a && foo(); // 42
+a && foo(); // 82
 ```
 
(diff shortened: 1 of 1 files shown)


Ivan-Perez:
42 * 2 = 84

;)

types & grammar/ch4.md
```diff
@@ -1234,15 +1234,15 @@ There's another idiom that is quite a bit less commonly authored manually, but w
...
 ```js
 function foo() {
-	console.log( a );
+	console.log( a * 2 );
 }
 
 var a = 42;
 
-a && foo(); // 42
+a && foo(); // 82
```
Ivan-Perez: 42 * 2 is 84

;)

PR #1026, opened 2017-05-10
modify isSafeInteger polyfill

DiligentYe:
__Number.MAX_SAFE_INTEGER__ is defined by ES6, so __Number.MAX_SAFE_INTEGER__ should be replaced by  __Math.pow(2, 53) - 1__ there.

### DIFF
--- a/types & grammar/ch2.md
+++ b/types & grammar/ch2.md
@@ -446,7 +446,7 @@ To polyfill `Number.isSafeInteger(..)` in pre-ES6 browsers:
 if (!Number.isSafeInteger) {
 	Number.isSafeInteger = function(num) {
 		return Number.isInteger( num ) &&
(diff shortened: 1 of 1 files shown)


legodude17:
Don't you mean `isSafeInteger` and not `isSaftInteger`?

DiligentYe:
Oh, sorry! I mean isSafeInteger!

legodude17:
That makes much more sense. 😄

getify:
The assumption (though not stated) is that this polyfill would have relied on `Number.MAX_SAFE_INTEGER` having been defined, too. I think most usages would have done that, so I think it's safe. Perhaps at most we just need a code comment mentioning that assumption, rather than duplicating `Math.pow(..)` logic.

PR #1034, opened 2017-05-24
Update ch6.md

WispProxy:
Small fix


getify:
This (original) style is from O'Reilly style guide stuff. It's a little annoying to me, but that's how they do it.

PR #1035, opened 2017-05-24
Declare vs Define

RickSanch3z:
Hello,
Throughout your book I found statements about functions declaration. How I understand more appropriate word is "definition". I just changed some "declare" to "define". I do not sure. Maybe I do not fully understand "declare" vs "define".
Thanks.

### DIFF
--- a/scope & closures/ch3.md
+++ b/scope & closures/ch3.md
@@ -7,7 +7,7 @@ But what exactly makes a new bubble? Is it only the function? Can other structur
 
 ## Scope From Functions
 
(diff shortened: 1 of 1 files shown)


tchan:
"Declare" is perhaps more technically correct than "define".
See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/function

Although I would go as far to say both are correct. *Shrug

pixilz:
I would also note in many languages the creation of a function doesn't instantiate anything, but in JavaScript it is immediately instantiated as an object.

kra3:
I think it make more sense in C example.

```c
int sum(int, int);    // declaration 
// after a long scroll - here you could use sum() 
int sum(int a, int b) { return a + b; }   // definition
```

I don't see such a different with JavaScript.

PR #1073, opened 2017-06-29
Refined syntactics

m4jing:


this & object prototypes/ch4.md
```diff
@@ -467,7 +467,7 @@ Another.count; // 1 (不会和`Something`共享状态)
 
 当一个传统的类被实例化时，就发生了类的行为向实例中拷贝。当类被继承时，也发生父类的行为向子类的拷贝。
 
-多态（在继承链的不同层级上拥有同名的不同函数）也许看起来意味着一个从子类回到父类的相对引用链接，但是它仍然只是拷贝行的的结果。
+多态（在继承链的不同层级上拥有同名的不同函数）也许看起来意味着一个从子类回到父类的相对引用链接，但它仍然只是拷贝的结果。
```
JoeHetfield: The original text:

> Polymorphism (having different functions at multiple levels of an inheritance chain with the same name) may seem like it implies a referential relative link from child back to parent, but it's still just a result of copy behavior.

so, I think the current translation is ok.

m4jing: 『拷贝行』是指什么？如果一定要忠实原文，『拷贝行为』貌似更合适

JoeHetfield: Yep, That's the real problem, I lost a 『为』:(

m4jing:
@JoeHetfield Fixed

JoeHetfield:
All right, thanks!

PR #1075, opened 2017-06-30
Optimization for accuracy

m4jing:
Here, the point is `getter` and `setter`.
Because the `getter` is hard-coded, so the `setter` would be ignored.

Please check, thanks.


this & object prototypes/ch3.md
```diff
@@ -675,7 +675,7 @@ myObject.a = 3;
 myObject.a; // 2
 ```
 
-因为我们仅为`a`定义了一个getter，如果之后我们试着设置`a`的值，赋值操作并不会抛出错误而是无声地将赋值废弃。就算这里有一个合法的setter，我们的自定义getter将返回值硬编码为仅返回`2`，所以赋值操作是没有意义的。
+因为我们仅为`a`定义了一个getter，如果之后我们试着设置`a`的值，赋值操作并不会抛出错误而是无声地将赋值废弃。就算这里有一个合法的setter，我们的自定义getter将返回值硬编码为仅返回`2`，所以`set`操作是没有意义的。
```
JoeHetfield: The original text is:

> Since we only defined a getter for a, if we try to set the value of a later, the set operation won't throw an error but will just silently throw the assignment away. Even if there was a valid setter, our custom getter is hard-coded to return only 2, so the set operation would be moot.

Your way has no problem, but I think we'd better follow the author's mind.

m4jing: 我在描述里也写了：

>Here, the point is `getter` and `setter`.
Because the `getter` is hard-coded, so the `setter` would be ignored.

这部分讲的是 `getter` 和 `setter`；

原文意思是：`getter`已经硬编码为就是返回2，那么`setter`再怎么写也不会影响`[[Get]]`时的取值了。而且，这里的 『set』，指的是`setter`的行为，而不是赋值。

这就是『the author's mind』, 所以我说的是『Optimization for accuracy』；谢谢

JoeHetfield: OK, I agree. But what I want to do here is faithfully translate author's writing. so, forgive me I insist this way.

m4jing: So, what's the meaning of the word 'set' here?

JoeHetfield: It means this line in the code snippet:

```js
var myObject = {
	// define a getter for `a`
	get a() {
		return 2;
	}
};

myObject.a = 3;  // <-- here is the 'set operation'

myObject.a; // 2
```

So, I think translate as '赋值操作' is fair enough.

m4jing: Yes.
```
myObject.a = 3;  // <-- here is the 'set operation'
```
Assignment?
No, it's in fact `set` operation.

It's not a serious problem though, just for accuracy.
You can either keep it or accept my suggestion.

JoeHetfield: OK, Sorry for I insist my opinion

m4jing: Up to you. 😯

JoeHetfield:
Ok, I close this one.

PR #1090, opened 2017-07-13
Update for ch5.md (Section - Multiple Views)

ghost:
If we try using of "buf" = 2 bits, we will have exceptions for all variables.

**Browser:** Chrome latest

**Exception:**
=========================
VM200:2 Uncaught RangeError: Invalid typed array length: 2
    at typedArrayConstructByArrayBuffer (<anonymous>)
    at new Uint16Array (native)
    at <anonymous>:2:13

**Valid example:**
=========================
var buf1 = new ArrayBuffer( 4 );
var first = new Uint16Array( buf1, 0, 2 )[0];

var buf2 = new ArrayBuffer( 3 );
var second = new Uint8Array( buf2, 2, 1 )[0];

var buf3 = new ArrayBuffer( 4 );
var third = new Uint8Array( buf3, 3, 1 )[0];

var buf4 = new ArrayBuffer( 20 );
var fourth = new Float32Array( buf4, 4, 4 )[0];

=========================
Need add to example "buf" = 20 or set custom "buf" for each variable.
After that please update the text if this remark is correct.

### DIFF
--- a/es6 & beyond/ch5.md
+++ b/es6 & beyond/ch5.md
@@ -98,6 +98,8 @@ If the buffer of binary data includes data in non-uniform size/location, this te
 For example, consider a binary buffer that has a 2-byte number (aka "word") at the beginning, followed by two 1-byte numbers, followed by a 32-bit floating point number. Here's how you can access that data with multiple views on the same buffer, offsets, and lengths:
 
 ```js
+var buf = new ArrayBuffer( 20 );
+
 var first = new Uint16Array( buf, 0, 2 )[0],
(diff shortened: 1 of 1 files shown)


getify:
It's true that `buf` actually has to be an array... and the snippet doesn't show that. But the paragraph above explains that you have to assume `buf` already has specific contents. Your addition only makes an empty `buf` which actually conflicts with what the paragraph says.

A perhaps clearer approach would be to have a code comment in the snippet (in place of that `buf` declaration) that reinforces that `buf` is assumed to already exist as a buffer.

Issue #1096, opened 2017-07-23
Chinese translation PR


hijiangtao:
Hey, thanks really a lot for contributing these series of books, which make me understand JavaScript a lot . Here are some typing faults when I read chinese translations (Book **this 与对象原型** - 第六章: 行为委托):

Words such as "面相对象" should be "面向对象" exactly, and I believe "然" in some places should be "让" instead. I fix these places in a pull request #1097 .

PR #1097, opened 2017-07-23
Refined part of Towards Delegation-Oriented Design

hijiangtao:
Details are in ISSUE #1096


this & object prototypes/ch6.md
```diff
@@ -29,9 +29,9 @@
 
 比方说我们有几个相似的任务（“XYZ”，“ABC”，等）需要在我们的软件中建模。
 
-使用类，你设计这个场景的方式是：定义一个泛化的父类（基类）比如`Task`，为所有的“同类”任务定义共享的行为。然后，你定义子类`XYZ`和`ABC`，它们都继承自`Task`，每个都分别添加了特化的行为来处理各自的任务。
+通过类，你可以通过这种方式设计这个场景：定义一个泛化的父类（基类）比如`Task`，为所有的“同类”任务定义共享的行为。然后，你定义子类`XYZ`和`ABC`，它们都继承自`Task`，每个都分别添加了特化的行为来处理各自的任务。
```
JoeHetfield: The original text is:

> With classes, the way you design the scenario is: define a general parent (base) class like `Task`, defining shared behavior for all the "alike" tasks. Then, you define child classes `XYZ` and `ABC`, both of which inherit from `Task`, and each of which adds specialized behavior to handle their respective tasks.

So I think the translation is just OK, and there are tow '通过' in your version, which I think is not a very fluent expression.

this & object prototypes/ch6.md
```diff
@@ -29,9 +29,9 @@
...
 
-**重要的是，** 类设计模式将鼓励你发挥继承的最大功效，当你在`XYZ`任务中覆盖`Task`的某些泛化方法的定义时，你将会想利用方法覆盖（和多态），也许会利用`super`来调用这个方法泛化版本，为它添加更多的行为。**你很可能会找到几个可以“抽象”到父类中，或在子类中特化（覆盖）的地方**。
+**重要的是，** 类设计模式将鼓励你发挥继承的最大功效，当你在`XYZ`任务中覆盖`Task`的某些泛化方法的定义时，你将会想利用方法覆盖（和多态），也许会利用`super`来调用这个方法泛化版本，为它添加更多的行为。**你很可能会找到几个可以“抽象”到父类中的通用方法，或在子类中特化（覆盖）的地方**。
```
JoeHetfield: The original text: 

> **Importantly,** the class design pattern will encourage you that to get the most out of inheritance, you will want to employ method overriding (and polymorphism), where you override the definition of some general `Task` method in your `XYZ` task, perhaps even making use of `super` to call to the base version of that method while adding more behavior to it. **You'll likely find quite a few places** where you can "abstract" out general behavior to the parent class and specialize (override) it in your child classes.

It didn't mentioned anything about '通用方法', but instead just a word of 'place'. So, I think the translation is OK.

hijiangtao: I think "general behavior" refers to "通用方法" or "通用行为".

this & object prototypes/ch6.md
hijiangtao: Sorry for didn't review the modification of inappropriate grammar.

this & object prototypes/ch6.md
JoeHetfield: All right, This part of the text is talking about the OO theory, and you can see the word 'specialize', which is opposite of the 'general'. So I think the best word should be '一般化的' or '泛华的'. '通用' sounds more like 'common'. 

I think this would be better:

> 你很可能会找到几个这样的地方：可以“抽象”到父类中，并在子类中特化（覆盖）的一般化行为。

hijiangtao: All right, that sounds fine.

JoeHetfield:
Thanks for your help!

hijiangtao:
Thanks for your suggestion! :-)

PR #1098, opened 2017-07-24
1ed-zh-CN 修正错别字

fomenyesu:

In  async & performance/ch3.md : 

 1 ‘但Promise不仅是是一个单步的’ 中 ‘不仅是是’ 应当为 ‘不仅仅是’。
 2 '实际上一种反模式' 中的 ‘实际’ 应当为 ‘实现’。

Thank you very much for your wonderful book!


async & performance/ch3.md
```diff
@@ -2090,7 +2090,7 @@ p.then( function(){
 
 当然，这里的问题是：这些关于性能的微小零头的潜在疏忽，和我们在本章通篇阐述的Promise带来的益处相比，*还值得考虑吗？*
 
-我的观点是，在几乎所有你可能认为Promise的性能慢到了需要被考虑的情况下，完全回避Promise并将它的可靠性和组合性优化掉，实际上一种反模式。
+我的观点是，在几乎所有你可能认为Promise的性能慢到了需要被考虑的情况下，完全回避Promise并将它的可靠性和组合性优化掉，实现上一种反模式。
```
JoeHetfield: The original text:

> My take is that in virtually all cases where you might think Promise performance is slow enough to be concerned, it's actually an anti-pattern to optimize away the benefits of Promise trustability and composability by avoiding them altogether.

The 'it's actually an anti-pattern' phase was translate as '实际上一种反模式', I think the word usage is fine. But I did missed a '是', so the correct expression should be: '实际上是一种反模式'.

PR #1106, opened 2017-08-05
Update ch5.md Line 472

cssjunkie:
I've had no problem following the book, but on line 472 there is a syntax construction that has me completely baffled:

    deps[i] = modules[deps[i]];

I checked and this works, but I can't figure out exactly what it does or even why it's legal. It seems to be referencing a non-existent modules property, which should assign undefined to that slot in deps. I don't see how making all items in deps undefined does anything useful.  

Could you put in a word about what it's doing and how, please? :-)

### DIFF
--- a/scope & closures/ch5.md
+++ b/scope & closures/ch5.md
@@ -470,6 +470,7 @@ var MyModules = (function Manager() {
 	function define(name, deps, impl) {
 		for (var i=0; i<deps.length; i++) {
 			deps[i] = modules[deps[i]];
+			// above line is confusing me.
 		}
 		modules[name] = impl.apply( impl, deps );
 	}


getify:
deps[i] comes in as a string of the name of the module dependency. The assignment replaces that string with the actual module that's stored at that string location in the modules object.

cssjunkie:
deps[i] comes in as a string of the name of the module dependency. The assignment replaces that string with the actual module that's stored at that string location in the modules object.

-------------------------------------------


Yes, but it appears that the modules object is empty. At least I see no place where any properties get assigned to it. I do understand all the remainder of this code and I have tried hard to find where that might be happening. It MUST be happening, but I just don't see it.

John

getify:
modules[..] is assigned on the next line. So you define a module first (ostensibly with no dependencies) then the next module can reference it as a dependency.

cssjunkie:
I finally got it. Somehow I had gotten the idea that 
it was only the module name that was passing in 
thru the argument list, rather than the module itself. 

Sorry for bugging you about it, Kyle.

John

getify:
Only the module name (string) *does* come in via `deps`... But the entries in `deps` get replaced with their actual module definitions (having been defined in previous calls to `define(..)` and added to `modules[..]`). So `deps` becomes an array of the actual module definitions, so that they're all available to pass in to the current module's initialization.

cssjunkie:
Only the module name (string) does come in via deps... But the entries in deps get replaced with their actual module definitions (having been defined in previous calls to define(..) and added to modules[..]). So deps becomes an array of the actual module definitions, so that they're all available to pass in to the current module's initialization.
----------------------------------------


When I finally realized that, it all came clear. Took a while tho, oy.

John

PR #1111, opened 2017-08-09
ES6 & Beyond (ch7.md): Broken links on Regex section

awareness481:
Links pointing to _people.mozilla.org_ are currently being used. However, _people.mozilla.org_ has been discontinued and is offline.
The links were pointing to ES6/ES2015 sections. I updated the links to use **ecma-international.org**. Each link should point to its corresponding section on the specification


getify:
Thanks!

PR #1112, opened 2017-08-15
a tiro's little problem

zhaoxieluoke:
First ,thank you all. 
This series of books are awesome.I have already read most of them. 
But I may find a problem in  Branch 1ed-zh-CN / es6 & beyond / ch2.md , I think It is a conceptual problem, this 
line 640-----在`{ x: aa, y: bb }`这一行中，`x`和`y`代表对象属性。在`{ x: AA, y: BB }`这一行，`x`和`y` *也* 代表对象属性。-----
should change to
-----在`{ x: aa, y: bb }`这一行中，`x`和`y`代表对象属性。在`{ x: AA, y: BB }`这一行，`AA`和`BB` *也* 代表对象属性。-----
we always console the target not the source  (我们打印的是属性名,而不是属性值 对么? Maybe I am wrong, forgive me ,I am a tiro)


JoeHetfield:
Ah... This is not an issue.  The destructure of an object is actually a kind of action pattern for instruct the JS engine.
```
var     { x: AA, y: BB } = o;
```
The code above means take the value of  'x' and 'y' property from object 'o', and assign them to 'AA' and 'BB', respectively. so that's why in the text it say '在{ x: AA, y: BB }这一行，x和y 也 代表对象属性'.

zhaoxieluoke:
console.log(AA, BB) // 10 20
console.log(x, y) //x is not defined

I think "AA" and "BB" are the property(属性),'x' and 'y' are the value (属性值), is this wrong?

JoeHetfield:
As mentioned above, the "AA" and "BB" actually are the variables that accept the value which been hold by the property "x" and "y". That's why you can print them out. but for x and y, they only exist on the "o" object. so when you refer them without "o." prefix, you got an error.

zhaoxieluoke:
Think for while, I think I may understand what you said, thanks for your patience

JoeHetfield:
Yes. and the targets are "AA" and "BB"

zhaoxieluoke:
I get this, old tie(with hands).

PR #1119, opened 2017-08-19
修复链接错误使用中文括号问题

hijiangtao:


JoeHetfield:
The original text: 

>  One such polyfill is "Native Promise Only" (http://github.com/getify/native-promise-only), which I wrote!

It seems  the author didn't mean to mark this part as an link but as plain text, so I think we should follow his mind.

JoeHetfield:
Ah, just like your another PR, the double-byte parentheses was a mistake indeed.

hijiangtao:
Well, it seems I did a wrong modification, but it also needs another PR to fix its link mixed with both url and its followed chinese characters.

JoeHetfield:
Reopened, you can use this PR

JoeHetfield:
Use single-byte parentheses man, not add the new angle brackets

hijiangtao:
I don't know what you really want to express? Single-byte parentheses will not separate words exactly, such as "我写的！" with its previous sentence, just like below:

image

But only angle bracket can make a proper link without title set to it. I think the problem is to fix the impropriate linking, rather than replace the double-byte parentheses (in chinese version).

JoeHetfield:
OK，I got you, I think this is the way that github interpret  markdown. I tried single-byte parentheses on VSCode, it display properly：
image
so, how about just leave it, cause every similar places gonna have this issue, I will fix them when I found a better way.

hijiangtao:
Yes, it really does. I think it's VSCode that help fix the display problem instead, because everything followed by a url without space should be thought to a part of url itself. Similar problems occured in some other places, hope you get a better hint and "fix" them. Thank you.

PR #1121, opened 2017-08-19
Update ch3.md

hijiangtao:
链接被延长排版


JoeHetfield:
Thanks for your help, but I think the true suspects are the double-byte parentheses. pls PR again if you fixed it

hijiangtao:
Well, I don't think this is the author's original idea, check the screenshots as below:

image

image

JoeHetfield:
Yes, the double-byte parentheses was a mistake, I should use the single-byte parentheses.

hijiangtao:
A same problem with #1119

Issue #470, opened 2015-06-18
"this & object prototypes": ch2 issue about polyfill of Function.prototype.bind()


zxfwinder:
The snippet in the polyfill of Function.prototype.bind() :

```
this instanceof fNOP && oThis 
? this : oThis
```

I cannot really understand why still need `&& oThis` even though `this instanceof fNOP` already indicate that the function invoked with `new` operater.

I try this polyfill with below snippet

```
function foo(p1,p2) {
        this.val = p1 + p2;
}
var bar = foo.bind( null, "p1" );
var baz = new bar( "p2" );
baz.val; // p1p2
```

but the result is `undefined`  :(
The correct result can appear once I comment `&& oThis`.

zxfwinder:
any help?

getify:
TBH, I'm not sure I know why it was there or not. Just spent a few minutes checking the spec and I'm not really clear on how the check should happen. But:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind$compare?locale=en-US&to=763553&from=747163

Someone removed it from the polyfill. So maybe it wasn't supposed to be there. Shrugs. I kinda doubt it matters, fwiw.

getify:
update polyfill for second edition.

injune1123:
I think the problem lies in how the polyfill is used in the book.
```function foo(p1,p2) {
        this.val = p1 + p2;
}
var bar = foo.bind( null, "p1" );
var baz = new bar( "p2" );
baz.val; // p1p2
```
For me, the result is also undefined, as is mentioned by @zxfwinder 
It happens since foo is bind with `null`.  
```var bar = foo.bind( null, "p1" );``` 
If foo is bound with an empty object, it works as expected. 
Here is my revision: 
```function foo(p1,p2) {
        this.val = p1 + p2;
}
var dummyObj= {}
var bar = foo.bind( dummyObj, "p1" );
var baz = new bar( "p2" );
dummyObj.val //unfined
baz.val; // p1p2
```

PR #1137, opened 2017-09-06
maybe `&& oThis` part of check should be removed

AnatoliiKolesnyk:
For example, if `bind` call was like `foo = foo.bind(null, 1)`
And then we construct new object like `new foo(2)`
In this case `this instanceof fNOP && oThis` evaluates to false and we will use 'oThis` as `this` binding instead of newly created object

**This is why you have the error mentioned in https://github.com/getify/You-Dont-Know-JS/issues/470**

I'm sorry for creating PR without reading Contributions Guidelines, but I've just ran into this typo and decided to help :)

And then I saw that issue is already reported

### DIFF
--- a/this & object prototypes/ch2.md
+++ b/this & object prototypes/ch2.md
@@ -537,8 +537,7 @@ if (!Function.prototype.bind) {
 			fBound = function(){
 				return fToBind.apply(
 					(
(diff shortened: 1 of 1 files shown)


getify:
It's already set for addressing in second edition. Thanks.

PR #1139, opened 2017-09-10
Fix minor grammar issues in chapter 3 of Async & Performance

miqid:
Hello,

Just offering suggestions for some minor grammar issues I came across while reading this chapter 3 of the Async & Performance book. Specifically,

- Rewording: `which more has our attention at present` to `which has more of our attention at present`
- Hyphen removal to be consistent with word usage: `reason-able` to `reasonable`


getify:
The first change is slight but *reasonable*. However, the second change shifts meaning away from intent. "reasonable" commonly means "acceptable", but "reason-able" means "able to be reasoned about".

miqid:
@getify, understood—I've redacted that second change.

kricketune:
The word _promisory_ should also be changed to _promissory_.

getify:
There's already a logged issue to handle the difference in spelling of promisory vs promissory.

PR #1147, opened 2017-09-15
minor typo: remove 'be' from 'be become'

scraggo:
After removing the typo, the latter part of the sentence is still a little unclear: if either form works, then will most developers prefer the shorter form because it's shorter? Or is it better in some other way?


getify:
This *is* a typo, and *be* should be removed. Thanks.

PR #1150, opened 2017-09-21
Ch5 This & Objects typo: **and it's not marked as read-only (`writable:false`)** ==> (`writable:true`).md

elloo:
Edited a detail in line 97 for clarity. 

Changed "writable:false" to "writable:true" - "**and it's not marked as read-only (`writable:true`)**". 

This clashed with line 98 "...but it's marked as **read-only (`writable:false`)**".


getify:
Already reported many times, please search issues. Thanks.

PR #1176, opened 2017-11-02
Small fix on implicit relation between Strong and Static typing

leostera:
How strict is the typing (strong/weak) is orthogonal to when is the
type information available (static/runtime).

So saying that a strong typed language is static is not necessarily true.

Samples:
- Erlang is Strong-Dynamically typed
- C is Weak-Static typed (because of void*)

This might spread misinformation.

### DIFF
--- a/types & grammar/ch1.md
+++ b/types & grammar/ch1.md
@@ -7,7 +7,7 @@ Most developers would say that a dynamic language (like JS) does not have *types
 >
 > An ECMAScript language type corresponds to values that are directly manipulated by an ECMAScript programmer using the ECMAScript language. The ECMAScript language types are Undefined, Null, Boolean, String, Number, and Object.
 
-Now, if you're a fan of strongly typed (statically typed) languages, you may object to this usage of the word "type." In those languages, "type" means a whole lot *more* than it does here in JS.
+Now, if you're a fan of strongly typed languages, you may object to this usage of the word "type." In those languages, "type" means a whole lot *more* than it does here in JS.
 
 Some people say JS shouldn't claim to have "types," and they should instead be called "tags" or perhaps "subtypes".
 


odykyi (approved):

roboogle:
In particular, [https://github.com/getify/You-Dont-Know-JS/blob/master/up%20%26%20going/ch1.md#variables](here) it is (wrongly) said (bold is mine):

> Other languages emphasize types for values instead of variables. Weak typing, **otherwise known as dynamic typing**, allows a variable to hold any type of value at any time.

This statement should be changed e.g. to:
> Dynamic typing allows a variable to hold any type of value at any time.

getify:
The delicate balance here is that the up & going book is for complete beginners (either to JS or even to programming). They won't know or care about any of those nuanced differences. At best, they may have heard those terms thrown around and conflated. So I don't want to actually cover all the nuances, but I do want to use various terms they may have heard to broadly and loosely tie together.

OvermindDL1:
> They won't know or care about any of those nuanced differences.

These are pretty big and important differences though...

* Static/Strong: OCaml/Haskell
* Static/Weak:  C++/Java/C# (in most ways, they are kind of a mix)
* Dynamic/Strong:  Python/Erlang
* Dynamic/Weak:  Javascript/LUA

Essentially static means known at compile time, dynamic means known only at runtime, and strong means a single thing only is a specific type, and weak means that it could potentially be many types depending on how it is treated (also can happen with unbounded coercion in strongly-typed languages, which makes them not so strong then...).

I made a blog post a while back about the differences because I kept getting bugged at people getting them wrong...  ^.^;

http://blog.overminddl1.com/posts/programming-language-differences-compilation/

getify:
> These are pretty big and important differences though...

That kind of statement is not wrong, but it's not correct when made without context of audience. Of course those differences are very important, but only to some audiences. I would strongly assert that the audience for this book (beginners) don't and can't care about such details.

OvermindDL1:
> That kind of statement is not wrong, but it's not correct when made without context of audience. Of course those differences are very important, but only to some audiences. I would strongly assert that the audience for this book (beginners) don't and can't care about such details.

Better then not to say anything about it at all because otherwise:

> So saying that a strong typed language is static is not necessarily true.

It is better not to outright lie then.  ^.^

ostera:
Ping. Can we merge or close this? I'm doing some PR cleanup on my end 😃

getify:
The second editions are still pending and I was keeping this issue open with the "second edition" label to ensure it gets handled when that time comes. If you want to close it, that's fine. I'll still search for the labels, open or closed, when the time comes.

getify:
Closing as not applicable anymore, since the new book "Get Started" (2nd-ed) has replaced "Up & Going".

PR #1182, opened 2017-11-14
possible typo on ch5 in this & Object Prototypes

limabeans:
not read-only seems to imply writable: true


getify:
Not a typo, please search closed issues for discussion on this topic.

PR #1185, opened 2017-11-16
Fix formatting for poem

victor-homyakov:
Unfortunately citation breaks poem formatting, so I've wrapped poem in a fenced code block

### DIFF
--- a/types & grammar/apA.md
+++ b/types & grammar/apA.md
@@ -313,21 +313,23 @@ Keywords are the obvious ones like `function` and `switch`. Future reserved word
 
 StackOverflow user "art4theSould" creatively worked all these reserved words into a fun little poem (http://stackoverflow.com/questions/26255/reserved-keywords-in-javascript/12114140#12114140):
 
-> Let this long package float,
-> Goto private class if short.
-> While protected with debugger case,
-> Continue volatile interface.
-> Instanceof super synchronized throw,
-> Extends final export throws.
->
-> Try import double enum?
-> - False, boolean, abstract function,
-> Implements typeof transient break!
-> Void static, default do,
-> Switch int native new.
-> Else, delete null public var
-> In return for const, true, char
-> …Finally catch byte.
+```
+Let this long package float,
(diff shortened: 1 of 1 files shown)


getify:
It's a long quote, so I think blockquote (as currently formatted) is the most correct semantic, right? It's not code, so why format it with a code fence? Only to get fixed-width font (in some readers)?

victor-homyakov:
The problem is, with blockquote poem is not formatted properly: it's rendered as a single long string. The only way I've found so far to preserve original formatting is to use code fence.

getify:
Ahhh.. I never spotted that formatting issue in markdown before (as the printed book formatting was corrected).

I think the better "fix" is to have a new blank quote line in between each, like this:

```
> Let this long package float,
>
> Goto private class if short.
>
> While protected with debugger case,
>
> Continue volatile interface.
>
> Instanceof super synchronized throw,
>
> Extends final export throws.
>
> &nbsp;
>
> Try import double enum?
>
> - False, boolean, abstract function,
>
> Implements typeof transient break!
>
> Void static, default do,
>
> Switch int native new.
>
> Else, delete null public var
>
> In return for const, true, char
>
> …Finally catch byte.
```

Which renders in markdown as:

> Let this long package float,
>
> Goto private class if short.
>
> While protected with debugger case,
>
> Continue volatile interface.
>
> Instanceof super synchronized throw,
>
> Extends final export throws.
>
> &nbsp;
>
> Try import double enum?
>
> - False, boolean, abstract function,
>
> Implements typeof transient break!
>
> Void static, default do,
>
> Switch int native new.
>
> Else, delete null public var
>
> In return for const, true, char
>
> …Finally catch byte.

victor-homyakov:
Formatting for this fragment still looks a bit weird

> Try import double enum?
>
> - False, boolean, abstract function,
>
> Implements typeof transient break!
>

Compare it with original

```
Try import double enum?
- False, boolean, abstract function,
Implements typeof transient break!
```

PR #1191, opened 2017-11-26
Avoid use of "obviously" in Types & Grammar intro

Krinkle:
The word `obvious` and `obviously` aren't always bad to use. But as part of an explanation, I find that is is almost always out of place, given the intended audience may very well not know (and expectedly so).

I think its removal in this case improves the content for everyone.

See also: <https://css-tricks.com/words-avoid-educational-writing/>


getify:
Very reasonable suggestion, thanks!

PR #1201, opened 2017-12-17
added ACCESSORY_PRICE to the calculation of affordability

esthor:
In the base/original: If the 'bank_balance' variable were less than 209.97 but greater than 199.98, the old expression would evaluate and claim we could afford the extra purchase, but we wouldn't actually be able to afford it because the bank balance would be less than the 'amount' + 'ACCESSORY_PRICE'.

In my PR: affordability is determined by adding both the 'amount' variable and 'ACCESSORY_PRICE' constant and then checking if that total is less than the 'bank_balance' variable.

This just bothered me when I read it and imagined the overdraft fees in the breaking test cases for some poor chap. Hope I'm not being too nit-picky or totally missing something.

### DIFF
--- a/up & going/ch1.md
+++ b/up & going/ch1.md
@@ -477,7 +477,7 @@ var amount = 99.99;
 amount = amount * 2;
 
 // can we afford the extra purchase?
-if ( amount < bank_balance ) {
+if ( amount + ACCESSORY_PRICE < bank_balance ) {
 	console.log( "I'll take the accessory!" );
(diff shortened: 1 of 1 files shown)


getify:
I simplified the presented code example in the book from the full normally expected version, since this book is for brand-new beginners. Unfortunately that created a "domain" confusion where some people struggle to suspend reality and understand the code concepts only. :)

I'm already planning to address this for the second edition. Thanks!

PR #1203, opened 2017-12-20
Fix typo

arun1595:
Removes the stray 'T' character from 'Tread'


getify:
not a stray "t", the word is tread: http://www.dictionary.com/browse/tread?s=t

arun1595:
@getify Stupid me, I should've googled instead of assuming it was a typo.

Thanks for educating me @getify :smile:

PR #1205, opened 2017-12-24
Update non-readonly property descriptor value to `writeable:true`

Huachao:
Update the non-readonly property descriptor value to `writeable:true` in property chain


getify:
Please see existing issues.

PR #1219, opened 2018-01-23
www.es6fiddle.net dose not work

mhdalmajid:
replace the old link http://www.es6fiddle.net/ with  https://jsfiddle.net/


getify:
Their new site is "esfiddle.net". I was hoping they would add a redirect from the old domain.

PR #1225, opened 2018-01-31
Update Writable:false

Nikhilkapoor20:
This commit updates text from writable:false to writable:true


getify:
Duplicate. Please search for the many previous issues: https://github.com/getify/You-Dont-Know-JS/issues?utf8=%E2%9C%93&q=writable+false

PR #1226, opened 2018-02-05
IE has limited ES6 functionality

bruincub:
Edge has a more complete implementation of ES6 than IE so Edge should be recommended, not IE

[ES6 comaptibility table ](https://kangax.github.io/compat-table/es6/)


getify:
OK, thanks.

PR #1236, opened 2018-02-12
Fix broken links

keyfoxth:
Seem links can be parsed direct in Markdown when the front is in English but not Chinese.


JoeHetfield:
Thanks for the effort and help! but... I didn't get the point. These links looking good on the github page. I can't see any necessity to change them.

keyfoxth:
This is what the links shows in [github.com](https://github.com/getify/You-Dont-Know-JS/blob/1ed-zh-CN/async%20%26%20performance/apB.md).

snipaste20180213_103803

Like I said, the links can not be accurately parsed with use `()` when their previous word is Chinese or fullwidth characters. They will be parsed until the next English word or the end of the line.

 I think this is a bug of Github.

JoeHetfield:
Oh yes, you are right. And this is a known issue.
But how about add a half-width space after the right parenthesis? That's much simpler fix, and won't create too many differences with the original text.

keyfoxth:
I tested it, it worked successfully but need parentheses must be halfwidth too.

[markdown-link](https://github.com/keyfoxth/markdown-link)

But I do not think it is a good idea to use space between `)` and `，` or `。`.

keyfoxth:
I found another way to fix it. Use `<>` around the links.

Like this:
`我能吞下玻璃（<https://github.com/>）而不伤身体。`

It will be parsed:
我能吞下玻璃（<https://github.com/>）而不伤身体。

Maybe we can fix it in the master branch.

JoeHetfield:
em... actually, It should be a half-width space between any English words and Chinese characters, and I'm this doing right now (though not done yet). so I think the usage of half-width parenthesis followed by a half-width space is OK to me. 
Anyway, my basic idea is as faithful as possible to the original text. so, shall we go for the half-width parenthesis + half-width space way?

keyfoxth:
I used to think of changing all `()` to `（）` as much as possible. But it is not a simple matter and not suitable for batch processing. Maybe need to take a few hours.

If your primary concern is keeping consistency with the original text, not care fullwidth or halfwidth, then like you said, the best way is inserting a space after `)` unless change the original text.

My opinion now is to use `（<...>）`.

PR #1247, opened 2018-03-03
Removed unnecessary word

m4jing:
Removing the unnecessary word makes the sentence smooth.


JoeHetfield:
Nice catch! Thank you!

PR #1260, opened 2018-03-21
Fixed typos in Chinese

m4jing:


m4jing:
Aha! Just make it a better document, especially for new learners. 😃

PR #1262, opened 2018-03-22
Corrected example to use obj instead of window

oflisback:
The one-line diff is hopefully self-explanatory, the example should use obj and not window to work as intended.


getify:
This example is intentionally using `window` to show that it's being ignored because `bar(..)` is already hard-bound.

PR #1263, opened 2018-03-22
Remove unused variable in example

oflisback:
Removing the unused parameter to remove a potential source of confusion, a is such a popular symbol name in these examples .. :)


getify:
That variable is there on purpose, for symmetry with the other examples and also to emphasize, combatting a common misconception, that the `a` parameter has no overlap with the `this.a` reference.

PR #1270, opened 2018-04-10
Update ch1.md

cssjunkie:
The two example code blocks starting on lines 544 and 564 indicate the output ends at the number 9, when it actually ends at 10.


cssjunkie:
Oops, I was incorrect, NM.

Issue #1177, opened 2017-11-02
"this & Object Prototypes" Chapter 2: Is the 'a' parameter in the lexical this example needed?


enumjorge:
The first example illustrating arrow function lexical scope starts off like this:
```
function foo() {
	// return an arrow function
	return (a) => {
		// `this` here is lexically adopted from `foo()`
		console.log( this.a );
	};
}
```

The arrow function that foo returns takes a parameter called 'a'. Unless I'm missing something, I don't see that parameter being used anywhere in the example or the discussion that follows. What's the purpose of that parameter?

getify:
The `a` is only there to emphasize that `this.a` is not a lexical reference to a local variable `a`, but rather a property on the `this` object context.

a-ogilvie:
I don't think this carries the point very well. In this example, a is assigned to an object which contains its own a property - it's unnecessarily confusing.

PR #1294, opened 2018-05-27
Argument is not needed in arrow function

a-ogilvie:
Since the example is illustrating that the function takes its `this` from the enclosing scope, it doesn't need any arguments.


getify:
Please see #1177.

PR #1299, opened 2018-06-10
Corrected typo that indicated writable:false indicated not read-only …

TCW417:
…(page 88 in print version)

Noticed this while reading today. writable:false would indicate the property is read-only but the text reads "...it's not marked as read-only (writable:false)"  I'm pretty sure that should be writable:true.


getify:
Known issue, please search: https://github.com/getify/You-Dont-Know-JS/issues?utf8=%E2%9C%93&q=is%3Aissue+%22writable%3Afalse%22+

PR #1315, opened 2018-07-12
ch3: clarify what's being iterated over in a classic for loop

jdhines:
The previous statement was that the classic 'for' loop iterated over the array indices, but that's just coincidental due to setting the limit of i from 0 to the array's length. There's nothing special about that choice, as it could easily be written so that i goes from 0 to 1, or -5 to +5. Therefore, just added a slight clarification so that there's no implication that a regular for loop has anything particularly to do with iterating over an array's indices, just that it's often been used for such a purpose.

**Yes, I promise I've read the [Contributions Guidelines](hO)** (please feel free to remove this line).

### DIFF
--- a/this & object prototypes/ch3.md
+++ b/this & object prototypes/ch3.md
@@ -811,7 +811,7 @@ for (var i = 0; i < myArray.length; i++) {
 // 1 2 3
 ```
 
-This isn't iterating over the values, though, but iterating over the indices, where you then use the index to reference the value, as `myArray[i]`.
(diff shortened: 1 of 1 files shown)


getify:
I don't think the sentence in question is implying that **all** `for` loops iterate over an array's indices. The sentence starts with "This is iterating...", implying ***this*** specific previous snippet.

Had it said, "For loops iterate..." or "All for-loops iterate...", then I'd be inclined to agree that it would be misleading.

The prior context of this discussion is that `for..in` loops do in fact iterate over the keys/indices, and then I show a regular `for` loop that does the equivalent. That's all to contrast and setup the discussion of the `for..of` loop a few paragraphs later.

PR #1355, opened 2018-09-21
change `bar(..)` to `foo(..)`

ashishpatelcs:
the function name should be foo instead of bar in this explanation.


getify:
The text is correct as-is. As with other mentions of "bar" in that paragraph, the reference in question is to the call-site that looks like "bar()" and "new bar()", not to the underlying "foo" function that actually gets called.

PR #1358, opened 2018-09-27
It is recommended to change "的性能要高" to "消耗的性能要高"

GreatMagicianHC:
"但它不被允许是因为引擎的实现者发现，在设置时检查（并拒绝！）无限循环引用一次，要比每次你在一个对象上查询属性时都做相同检查的性能要高。"
这句话，目前翻译的意思为："无限循环引用"要比"每次在一个对象上查询属性"的性能要高。前者性能比后者好，为什么我们还要不允许使用呢？
回顾在英文版原文中提到："But it's disallowed because engine implementors have observed that it's more performant to check for (and reject!) the infinite circular reference once at set-time rather than needing to have the performance hit of that guard check every time you look-up a property on an object." 注意这里："that it's more performant to check" 个人理解应该翻译为消耗(花费)更多的性能，而不是性能更高。并且，综合文章上下文来看，相互委托不允许被使用，是引擎实现者发现的是"无限循环引用"比"在一个对象上查询"更消耗性能，所以才不允许使用。
综上，建议把"都做相同检查的性能要高"替换为"都做相同检查消耗的性能要高"


JoeHetfield:
Sorry man, I think the "more performant" means "more effective, faster and better", because the "performant" here is an adjective, not a noun. 

And think about this: 
check for the circular reference "once" when you setup a property
vs. 
check for circular reference "every time" when you use a property.

which one more effective? Obviously the former one.

PR #1367, opened 2018-10-23
Corrected output lines 36 and 37

noircir:
Lines 36 and 37 currently suggest that the reader should see an output KYLE and READER (as per comments). This output never happens, since the 'identify' function returns a value, but never logs it. I added "console.log", to avoid confusion for new readers.

### DIFF
--- a/this & object prototypes/ch1.md
+++ b/this & object prototypes/ch1.md
@@ -33,8 +33,8 @@ var you = {
 	name: "Reader"
 };
 
-identify.call( me ); // KYLE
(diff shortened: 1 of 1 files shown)


noircir:
This concerns the book _this_ & Object Prototypes, Ch. 1

getify:
It doesn't need a `console.log(..)` to display a value. But moreover, the code comments are suggesting the outputs, not necessarily the values printed.

screen shot 2018-10-23 at 7 06 38 pm

PR #1373, opened 2018-11-03
Correct an important phrase

bilgamers:
If I understood this lesson right, I think it should be TypeError.


themarketlise:
Thank you for your reply!

so now it means no matter what, the engine always performs a reference check.
when it's RHS, reference check then we get the value from that reference.
when it's LHS, reference check then we assign the value to that reference.

PR #1382, opened 2018-11-20
update to reflect the fact that console has spec

thefourtheye:
As `console.*` functions are supposed to follow the spec
https://console.spec.whatwg.org/, the text about it is
updated.

### DIFF
--- a/async & performance/ch1.md
+++ b/async & performance/ch1.md
@@ -96,9 +96,7 @@ Any time you wrap a portion of code into a `function` and specify that it should
 
 ### Async Console
 
-There is no specification or set of requirements around how the `console.*` methods work -- they are not officially part of JavaScript, but are instead added to JS by the *hosting environment* (see the *Types & Grammar* title of this book series).
-
-So, different browsers and JS environments do as they please, which can sometimes lead to confusing behavior.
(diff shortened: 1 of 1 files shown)


getify:
"supposed to follow the spec" is a bit of a stretch. This spec is not even complete, very early draft, and not even in conformance yet with all browsers. Maybe a few years from now it'll be the official standard, but certainly not yet.

getify:
It was entirely true at the time of writing. It's partially true still. It's worth potentially re-visiting in second edition, but I'm not convinced that we'll be at a point where that spec is having any influence on the browsers' behavior. Right now, they mostly do whatever they want, in service of developers. Having worked on a devtools team before, I can say that browsers tend to treat those tools as a way to differentiate from other browsers in the effort to attract developer mindshare. So I find it at least somewhat suspect that we'll ever be at a place where the console is entirely consistent according to a spec.

Right now, the spec's only stated goal is to define that which the consoles already do. It's not even complete at that, yet. But until browsers indicate strongly they plan to adhere to it, and importantly, to adhere to each others' standards, I'm not sure how close in spirit to a "spec" this document will ever be.

Always remember: specs are opt-in by implementations. A spec is only as good as participants choose to adhere to it.

async & performance/ch1.md
```diff
@@ -96,9 +96,7 @@ Any time you wrap a portion of code into a `function` and specify that it should
```
captain-hoity-toity, line 96: A Factory pattern might be a good fit for this instantiation logic.

IgorWnek (approved):
Nice 👍🏻

PR #1400, opened 2019-01-02
Update ch2.md

agate:
`typeof` can detect `function` as well.

### DIFF
--- a/up & going/ch2.md
+++ b/up & going/ch2.md
@@ -20,6 +20,7 @@ As we asserted in Chapter 1, JavaScript has typed values, not typed variables. T
 * `boolean`
 * `null` and `undefined`
 * `object`
+* `function`
 * `symbol` (new to ES6)
 
 JavaScript provides a `typeof` operator that can examine a value and tell you what type it is:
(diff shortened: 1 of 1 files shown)


getify:
First, the list you're referring to is not about the return values from `typeof`, it's about the actual defined types as per the spec. `function` is not a defined type (despite the fact that `typeof` returns it). Instead, the spec refers to a function as a sub-type of the `object` type, called a "callable object".

Secondly, the `typeof` operator only returns these 7 (not 8) string values:

1. `"undefined"`
2. `"string"`
3. `"boolean"`
4. `"number"`
5. `"object"`
6. `"function"`
7. `"symbol"` (as of ES6+)

There are 7 built-in types (`null`, `undefined`, `number`, `boolean`, `string`, `object`, `symbol`) plus the `function` object sub-type (similar to array), so it seems like `typeof` should have 8 return values. The reason it's only 7 is because `typeof` incorrectly (a long-standing bug) returns `"object"` for `null` instead of something sensible like `"null"`.

agate:
Aha. Thanks for the explanation! I see your point. Closing this PR.

PR #1431, opened 2019-03-03
Update ch1.md

deepakcdo:
Made it clear that `a` was requested while executing function foo

### DIFF
--- a/scope & closures/ch1.md
+++ b/scope & closures/ch1.md
@@ -159,7 +159,7 @@ Let's imagine the above exchange (which processes this code snippet) as a conver
 
 > ***Engine***: Great, thanks! OK, I'm executing `foo`.
 
-> ***Engine***: Hey, *Scope*, I've got an LHS reference for `a`, ever heard of it?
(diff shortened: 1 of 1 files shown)


scope & closures/ch1.md
```diff
@@ -159,7 +159,7 @@ Let's imagine the above exchange (which processes this code snippet) as a conver
 
 > ***Engine***: Great, thanks! OK, I'm executing `foo`.
 
-> ***Engine***: Hey, *Scope*, I've got an LHS reference for `a`, ever heard of it?
+> ***Engine***: Hey, *Scope*, While executing function `foo` I've got an LHS reference for `a`, ever heard of it?
```
getify: Is the "I'm executing Foo" in the line immediately above it not clear enough?

deepakcdo: When I first read it I was not full sure  what stage of exectuing function foo was at. When I read it the scond time I understood it. So I thought it would be clear to add that happens while execution the function.

PR #1437, opened 2019-03-13
Typo where I writeable should be true

deepakcdo:
Typo where I writeable should be true as it says 'it's not marked as read-only '

### DIFF
--- a/this & object prototypes/ch5.md
+++ b/this & object prototypes/ch5.md
@@ -94,7 +94,7 @@ If the property name `foo` ends up both on `myObject` itself and at a higher lev
 
 As we just hinted, shadowing `foo` on `myObject` is not as simple as it may seem. We will now examine three scenarios for the `myObject.foo = "bar"` assignment when `foo` is **not** already on `myObject` directly, but **is** at a higher level of `myObject`'s `[[Prototype]]` chain:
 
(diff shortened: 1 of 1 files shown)


getify:
Please search existing issues before filing.

deepakcdo:
I checked master before proposing this change.

Its difficult to check all the issues. If this has been reported and you agree this is an error how come the merge request was not accpected so mater is updated. 

If the errors are not merged into master then there is no point me providing my corrections.

getify:
> Its difficult to check all the issues

When you file a PR/issue there is a link to the contribution guidelines, which you had to remove as you were filing, indicating you read them. They explain, among other things, that it's better to search first, then file an issue, then file a PR last.

If you search for "writable", you'll find a dozen issues.

> how come the merge request was not accpected so mater is updated

First, I don't fully agree with the change, as is discussed in the various previous issues.

Second, this is a published book and I'm keeping master the same as the published 1st edition.

Also if you search there are over 100 issues (including writable) that are labeled "second edition". That work has not started yet (it's a lot to do, and there are some complications) so that's why that label is tracking all those issues.

Once second edition starts, master will be second edition and first edition will move to a branch.

deepakcdo:
Thanks for the note. 

Since most readers will read the book and want to check the errors quickly, is it worth having a branch which is the start of the second verison?  This is common practice for developers and will see what you are trying to achieve. 

So master can remain as you like ( published book).  

After searching for writeable like you suggested its clear many people are confused. So the second branch with some curicial fixes will help.

PR #1444, opened 2019-03-27
this & Object Prototypes: 2: function reference property

iskitz:
Proposal to rephrase “property function reference” to “function reference property” which more clearly communicates a function reference stored as a property.

The original phrasing can be confusing when read as “_property function_ reference” vs. “_function reference_ property”.

**Yes, I promise I've read the [Contributions Guidelines](https://github.com/getify/You-Dont-Know-JS/blob/master/CONTRIBUTING.md)**.

### DIFF
--- a/this & object prototypes/ch2.md
+++ b/this & object prototypes/ch2.md
@@ -233,9 +233,9 @@ Either way the `this` is changed unexpectedly, you are not really in control of 
 
 ### Explicit Binding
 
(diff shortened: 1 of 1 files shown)


getify:
Thanks for the suggestion!

iskitz:
👍🏾 you’re welcome. Thanks for all the knowledge in this excellent book.

Issue #1266, opened 2018-03-28
Async & Performance Chapter 3: Promises - ReferenceError instead of TypeError


iDanbo:
https://github.com/iDanbo/You-Dont-Know-JS/blob/master/async%20%26%20performance/ch3.md#swallowing-any-errorsexceptions

```javascript
var p = new Promise( function(resolve,reject){
	resolve( 42 );
} );

p.then(
	function fulfilled(msg){
		foo.bar();
		console.log( msg );	// never gets here :(
	},
	function rejected(err){
		// never gets here either :(
	}
);
```

I get a `ReferenceError: foo is not defined` instead of `TypeError`.  Is it a typo?
As I understood, the error will be caught anyway because of the spec changes, so running the code in Chrome will give `Uncaught (in promise) ReferenceError: foo is not defined`. But where is the `TypeError`?

getify:
You're correct, should have been `ReferenceError`. I think this was a missed detail when the code snippet was refactored from an earlier version. Will address in second edition.

PR #1468, opened 2019-05-15
Update async & performance/ch3.md (fix error type)

guzoff:
I think that the error type mentioned in text should be changed from TypeError to ReferenceError (or maybe some 'foo' variable should be defined in global or function scope to get actually TypeError).


getify:
Known issue #1266

PR #1474, opened 2019-06-02
Update ch2.md

devpurohit:
The **Modules** subsection of Chapter 2 in **Up & Going** has a code snippet which has an error in the function call. This PR resolves that.

### DIFF
--- a/up & going/ch2.md
+++ b/up & going/ch2.md
@@ -742,7 +742,7 @@ function User(){
 // create a `User` module instance
 var fred = User();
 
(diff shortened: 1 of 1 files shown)


up & going/ch2.md
```diff
@@ -742,7 +742,7 @@ function User(){
 // create a `User` module instance
 var fred = User();
 
-fred.login( "fred", "12Battery34!" );
+fred.doLogin( "fred", "12Battery34!" );
```
devpurohit: Correct function is being called now.

PR #1488, opened 2019-07-15
Changed confusing sentence that assumes object is a primitive

AmazingDesign:
ECMA spec says `A primitive value is a member of one of the following built-in types: Undefined, Null, Boolean, Number, String, and Symbol; an object is a member of the built-in type Object; and a function is a callable object.` so Object is not a primitive. Sentence changed in this commit was confusing.
I believe that there isn't anything like `[object Array]`. It can't be produced to by .toString() nethier .valueOf() on array, and spec does not contain it. Can't it be simply `Array`?


AmazingDesign:
Ok, I get it, `Object.prototype.toString.call( [] )` can produce `[object Array]`, but i still think that after change the sentence is less confusing.

Issue #1510, opened 2019-09-05
Scope & Closures: Chapter 2 Lexical Scope - Confusing Wording


AyoubIssaad:
**"I already searched for this issue":**

**Edition:** (2nd)

**Book Title:** Scope & Closures

**Chapter:** Lexical Scope

**Section Title:** Introduction: First Paragraph

**Problem:** 
The first paragraph reads as follows: 
> In Chapter 1, we defined "scope" as the set of rules that govern how the Engine can look up a variable by its identifier name and find it, either in the current Scope, **### or in any of the _Nested_ Scopes it's contained within**.

I think the last sentence is confusing because of the incorrect use of the term "Nested Scopes" because:

'Nested scopes' implies that the scopes are contained within the current scope, and it contradicts with the second part of the sentence "it's contained within".

I think the correct sentence would be:
"... either in the current Scope, or in any of the **OUTER** Scopes it's contained within."

Thank you

PR #1511, opened 2019-09-05
Outer scopes instead if Nested scopes

AyoubIssaad:
Issue Number:  #1510

"I already searched for this issue":

Edition: (2nd)

Book Title: Scope & Closures

Chapter: Lexical Scope

Section Title: Introduction: First Paragraph

Problem:
The first paragraph reads as follows:

In Chapter 1, we defined "scope" as the set of rules that govern how the Engine can look up a variable by its identifier name and find it, either in the current Scope, ### or in any of the Nested Scopes it's contained within.

I think the last sentence is confusing because of the incorrect use of the term "Nested Scopes" because:

'Nested Scopes' implies that the scopes are contained within the current scope, and it contradicts with the second part of the sentence "it's contained within".

I think the correct sentence would be:
"... either in the current Scope, or in any of the OUTER Scopes it's contained within."

Thank you

----

**Edition:** (1st or 2nd)

**Book Title:**

**Chapter:**

**Section Title:**

**Topic:**

### DIFF
--- a/scope-closures/ch2.md
+++ b/scope-closures/ch2.md
@@ -21,7 +21,7 @@
 
 ----
 
-In Chapter 1, we defined "scope" as the set of rules that govern how the *Engine* can look up a variable by its identifier name and find it, either in the current *Scope*, or in any of the *Nested Scopes* it's contained within.
+In Chapter 1, we defined "scope" as the set of rules that govern how the *Engine* can look up a variable by its identifier name and find it, either in the current *Scope*, or in any of the *Outer Scopes* it's contained within.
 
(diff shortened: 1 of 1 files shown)


getify:
The outer scopes are all nested within each other. Perhaps the phrase could have said "outer nested scopes" but I don't think either "nested scopes" or "outer scopes" is more or less accurate than the other.

AyoubIssaad:
Got that, thank you for clarifying,  I just want to add this

> The outer scopes are all nested within each other.
This might not be the case if the current scope is one level within the global scope, I don't think in this case the outer scope, which is the global one is nested.

PR #1522, opened 2019-10-17
fixed case letter, book es-next-beyond, chapter 2, line 518

alexsenichev:
----

**Edition:** 2nd

**Book Title:** es-next-beyond

**Chapter:** 2

**Section Title:** Default Parameter Values

**Topic:** Default Value Expressions


getify:
Thank you, but I don't need copy-editing on the previous version of the book that was published almost 4 years ago. All six books are being heavily revised for second edition, and will be copy-edited once they're draft complete.

PR #1526, opened 2019-10-26
2nd ed

pmzi:
1. In the **Operators `||` and `&&`** section, there is a sentence that says:
```
If the test is `false`, the `&&` expression results in the value of the *first operand* (`a` or `c`)
```
For example:
```javascript
console.log(true && '')
```
It will result in the second operand.

2. In the **Abstract Relational Comparison section** you said:
```
The algorithm first calls `ToPrimitive` coercion on both values, and if the return result of either call is not a `string`, then both values are coerced to `number` values using the `ToNumber` operation rules, and compared numerically
```
Which its example is wrong because they have been compared alphabetically:
```javascript
[42] < ["43"]
```
Because `[42]` will be "42" and after that "42" will be compared to "43".
As I explained:
```javascript
[42] < ["043"] // false
```
----

**I already searched for this issue**

**Edition:** 2nd

**Book Title:** You-Dont-Know-JS

**Chapter:** Chapter 4: Coercion

**Section Title:** types-grammar

**Topic:** Abstract Relational Comparison & Operators `||` and `&&`

### DIFF
--- a/types-grammar/ch4.md
+++ b/types-grammar/ch4.md
@@ -1200,7 +1200,7 @@ Both `||` and `&&` operators perform a `boolean` test on the **first operand** (
 
 For the `||` operator, if the test is `true`, the `||` expression results in the value of the *first operand* (`a` or `c`). If the test is `false`, the `||` expression results in the value of the *second operand* (`b`).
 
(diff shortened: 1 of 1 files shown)


getify:
This book hasn't been worked on for second edition yet.

pmzi:
Didn't understand, any plans to explain more about it?

getify:
The text you're talking about is from 1st edition. This specific 2nd edition book hasn't started yet. There's no reason to be submitting PRs for text that is "in progress" but that literally hasn't even started.

PR #1529, opened 2019-10-30
Update ch2.md

pranithsrujanroy:
the variable name is a block scoped variable but it's confusing when said it behaves as a variable declared with var

**Yes, I promise I've read the [Contributions Guidelines](https://github.com/getify/You-Dont-Know-JS/blob/master/CONTRIBUTING.md)** 

----

**Edition:** 2nd

**Book Title:** You Don't Know JS Yet

**Chapter:** 2

**Section Title:**

**Topic:**

### DIFF
--- a/getting-started/ch2.md
+++ b/getting-started/ch2.md
@@ -221,7 +221,7 @@ hello("Kyle");
 
 Here, the identifier `hello` is created in the outer scope, and it's also automatically associated so that it references the function. But the named parameter `name` is created only inside the function, and thus is only accessible inside that function's scope.
 
-Both `hello` and `name` act generally as if they were declared with `var`.
+`hello` acts as if it was declared with `var` where as `name` declares as if declared with `let`.
 
 Another syntax that declares a variable is the `catch` clause of a `try..catch` statement:
 


pranithsrujanroy:
It says - Both `hello` and `name` act generally as if they were declared with `var`.
but `name` being a function parameter is block scoped and behaves like a variable declared with `let`.

In the later example of `catch` block isn't it the similar case where `err` is a parameter where it is written is block scoped and behaves as if declared with `let`

getify:
> but name being a function parameter is block scoped and behaves like a variable declared with let.

This is not accurate. Parameters are available to the entire function, and thus act more like a `var`. Moreover, you can do `var paramName = ..` and "re-declare" a variable, unlike `let` which would disallow a re-declaration.

The way in which parameters do act more like a `let` declaration is if you're using `= ..` default parameter expressions, in which case you can violate the TDZ (which `var`s do not have). So parameters are sorta like both, but I would claim they are much more like `var`, especially historically, than like `var`.

pranithsrujanroy:
> This is not accurate. Parameters are available to the entire function, and thus act more like a `var`. Moreover, you can do `var paramName = ..` in the function body to "re-declare" a variable, unlike `let` which would disallow such a re-declaration.
> 
then isn't is the same with _`err`_  in catch block. I see the same scenario over there but there it is written it behaves more like a variable declared with`let`

PR #1555, opened 2019-12-02
2nd ed typos

parid0kht:
some typos for get started.


getify:
Thank you for these. Keep in mind that I have a copy-editor who will review all the books for typos, so stuff like this will be addressed before publishing.

parid0kht:
> Thank you for these. Keep in mind that I have a copy-editor who will review all the books for typos, so stuff like this will be addressed before publishing.

noted on this and to be honest, for some but not all of the typos I used a spell checker to find most of the typos for these 3 chapters before pushing.

PR #1560, opened 2019-12-07
Update ch3.md

sachinmaskalle:
Typo


getify:
Sift is the intended word.

PR #1566, opened 2019-12-13
typo: 'strict mide' should be 'strict mode' ?

ya-luotao:
should be 'strict mode' I guess?

**Edition:** 2nd-ed

**Book Title:** You Don't Know JS Yet

**Chapter:** Chapter 1: What Is JavaScript?

**Section Title:** Strictly Speaking

**Topic:**

### DIFF
--- a/get-started/ch1.md
+++ b/get-started/ch1.md
@@ -425,7 +425,7 @@ The strict mode pragma must appear at the top of a file, with only whitespace or
 
 | WARNING: |
 | :--- |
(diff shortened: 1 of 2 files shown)


getify:
mide should indeed be mode. but JS'y was intended. it means "JS-like".

ya-luotao:
> mide should indeed be mode. but JS'y was intended. it means "JS-like".

Thanks! OK, let me revert the JS'y change commit.

PR #1589, opened 2020-01-13
New features: Add gitbook feature( Read IT with gitbook )

magicianlib:
## Note

We didn't make any changes, Just a few more files (related to gitbook).

In order to be able to read it like a book, easier to read.

Now the reader just needs to clone the repository for download, and simply run a command in the repository to open the book in the browser.
There is a catalog with instructions, and you can also customize the gitbook plug-in, customized exquisite books.

I think everyone needs better reading habits.

## Change

Add Some file:

- `SUMMARY.md` Gitbook Summary
- `book.json` Gitbook Config(plugin)
- `styles/` gitbook style(css file)

## Sample

Read the cover(open it in a browser(`localhost:4000`)):

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,19 @@
+# Created by .ignore support plugin (hsz.mobi)
+### GitBook template
+# Node rules:
+## Grunt intermediate storage (http://gruntjs.com/creating-plugins#storing-task-files)
+.grunt
+
+## Dependency directory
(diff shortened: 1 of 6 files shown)


getify:
I'm not sure if I'm going to support gitbook or not, but I am definitely not going to do so while the books are still in progress. I will revisit the topic after they're complete.

PR #1592, opened 2020-01-22
Update chapter 2

Lafaa:
In the declaration of the class "Book" you declared the constructor to accept an Object with 3 properties (title, author, publishedOn), but then when you instanciate that class, you use the 3 values not wrapped into an object. Since in all the rest of the examples you did not use wrapping objects, I think it's best to correct the class declaration instead of fixing the "new Book(...)"  statement.

EDIT: The same is true for the Modules section

**Edition:** 2nd Edition

**Book Title:** Get Started

**Chapter:** Chapter 2

**Section Title:** How We Organize in JS

**Topic:** Classes & Modules

### DIFF
--- a/get-started/ch2.md
+++ b/get-started/ch2.md
@@ -603,11 +603,11 @@ Now let's consider more specific types of publication, like `Book` and `BlogPost
 
 ```js
 class Book extends Publication {
-    constructor(bookDetails) {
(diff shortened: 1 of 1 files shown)


getify:
This difference is on purpose, to illustrate that a child class can define its own constructor with a different kind of signature from the parent constructor.

PR #1593, opened 2020-01-22
multiple pages: fix typos

schneiderl:


schneiderl:
For the record, I think [ch1](scope-closures/ch1.md) changing log is not properly showing what has changed. The word "occurrence" was wrong there.

getify:
Thanks, I got these in.

kailando:
@getify *Why, oh *why** didn't you do it?

PR #1600, opened 2020-02-08
generator functions syntax correction

Mhmdrza:
the "*" should be stick to function keyword not the function name;
I already searched for this issue, reference:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*


getify:
The `*` can be placed either next to `function` or next to the function's name, with or without spaces on either side of it. It's a stylistic preference. I prefer to put it next to the function name.

PR #1601, opened 2020-02-14
fix reference to figure 1

devmi:
**Yes, I promise I've read the [Contributions Guidelines](https://github.com/getify/You-Dont-Know-JS/blob/master/CONTRIBUTING.md) and I already searched for this issue"**

**Edition:** 2nd

**Book Title: You Don't Know JS Yet: Get Started**

**Chapter: 1**

**Section Title: What's in an Interpretation?**

**Topic: is it [JS] compiled?**


getify:
Good catch on the figure reference, but the book was finished with its writing, and indeed published, in January.

PR #1613, opened 2020-02-25
Babel is making the code "backwards-compatible", right?

spidgorny:
It seems the terms got mixed. Please check.

> Typos?
>
>

----

**Book Title:**

**Chapter:**

**Section Title:**

**Topic:**


getify:
The text is correct as stated. Quoting from just a little bit above that:

> Being forwards-compatible means that including a new addition to the language in a program would not cause that program to break if it were run in an older JS engine.

So, Babel is helping fix that JS is **not, by itself** forwards-compatible, in that new additions to the language *would* break an older JS engine. But thankfully, Babel "fixes" that.

JS is, by its nature, backwards-compatible, in that older code still runs in newer engines already (with a few extremely minor and rare exceptions). We don't need any extra tooling for backwards-compat, it's already baked in to how JS is managed and evolved (the "One JS" principle).

PR #1647, opened 2020-03-24
modify: toggle example

albseb511:

> Typos?
>
>

----

**Book Title:** 

`scope-closures`

**Chapter:** 

`apB.md`

**Section Title:** 

`Suggested Solutions`

**Topic:**

`Suggested: Closure (PART 2)`

Modified the example for toggle. A bit more compact and uses arguments. I found it a bit more easier, as we are not doing any operations on the array like shift / unshift. 

```javascript
function toggle(){
    var i = 0
    var list = arguments
    return function(){
      if(i===list.length)
         i = 0
      return list[i++]    
    }
}
```

### DIFF
--- a/scope-closures/apB.md
+++ b/scope-closures/apB.md
@@ -468,19 +468,14 @@ I also renamed the inner function from `factorize(..)` to `findFactors(..)`. Tha
 The *Closure Exercise (PART 2)* `toggle(..)` can be solved like this:
 
 ```js
-function toggle(...vals) {
-    var unset = {};
-    var cur = unset;
-
-    return function next(){
-        // save previous value back at
-        // the end of the list
-        if (cur != unset) {
-            vals.push(cur);
-        }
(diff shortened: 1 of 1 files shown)


getify:
I am glad you've found your own solution that you like better -- that means the book did its job in teaching you! -- but I'm not going to update the book with each person's own approaches. I just presented **one** suggested solution. Thanks.

albseb511:
Yeah that's fair enough. Just thought of making the example better.

Would it be valuable if I can contribute to different examples in the
future? Rather than changing existing ones?

getify:
Once I publish a book (like once it's complete) it's unlikely that I'll change the book just with alternate approach suggestions unless there are legitimate defects. This book was published about a month ago, so it's pretty unlikely anything more than minor fixes will occur.

While I'm writing a book here, suggestions are welcomed.

PR #1649, opened 2020-03-27
Small Change in Chapter 4 (Types and Grammar)

aditya-mitra:
It is not just the second operand but the last operand which is returned when all the conditions are true for &&.
Similarly, it is not just the second operand but the first operand, which is true, that is returned for ||.

`var a =42;
var b = "abc";
var c = null;
var d = 1;
var e = 0;`
`a && b && d; //1`
`e ||  c ||  a; //42`

### DIFF
--- a/types-grammar/ch4.md
+++ b/types-grammar/ch4.md
@@ -1202,9 +1202,9 @@ c && b;		// null
 
 Both `||` and `&&` operators perform a `boolean` test on the **first operand** (`a` or `c`). If the operand is not already `boolean` (as it's not, here), a normal `ToBoolean` coercion occurs, so that the test can be performed.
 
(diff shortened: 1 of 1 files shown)


getify:
This book was published almost 5 years ago, so it's not subject to any changes at this point.

Moreover, you're changing the meaning of the text to refer to a different concept (operator associativity), which is dealt with elsewhere in the text.

For posterity, the current text uses "first" and "second" (operands) because the context of discussion is just one usage of `||` or `&&`. Your changes mix in the concept of multiple uses of the operator strung together. The way that works is the operators are still processed one at a time (so, still a "first" and "second" operand), and the order that the operators are handled is left-to-right because the `||` and `&&` are left-associative.

PR #1660, opened 2020-04-30
Adding missing words/ removing extra words.

buchs:

You may or may not classify these changes as minor. The earlier you kill kill bugs, the less it costs!


getify:
I appreciate the suggestions for wordings but I don't feel I want to make those preferential changes since the book has already been published and sold quite a few copies.

PR #1665, opened 2020-05-04
Fix grammar

henrebotha:
I already searched for this issue :upside_down_face:

### DIFF
--- a/get-started/ch2.md
+++ b/get-started/ch2.md
@@ -49,7 +49,7 @@ greeting("My name is Kyle.");
 
 In this program, the value `"My name is Kyle."` is a primitive string literal; strings are ordered collections of characters, usually used to represent words and sentences.
 
-I used the double-quote `"` character to *delimit* (surround, separate, define) the string value. But I could have used the single-quote `'` character as well. The choice of which quote character is entirely stylistic. The important thing, for code readability and maintainability sake, is to pick one and to use it consistently throughout the program.
+I used the double-quote `"` character to *delimit* (surround, separate, define) the string value. But I could have used the single-quote `'` character as well. The choice of which quote character is entirely stylistic. The important thing, for the sake of code readability and maintainability, is to pick one and to use it consistently throughout the program.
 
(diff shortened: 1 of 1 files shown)


getify:
Thanks for the suggestions. Not clear to me that these are strictly "fixes" as opposed to just preferences.

henrebotha:
These are indeed fixes. The expression is "for the sake of X" or "for X's sake". It is grammatically invalid to say, "I did this for clarity sake."

getify:
Do you have a reference for that assertion? I ask because I've had the books copy-edited and the editor did not assert this rule to make such corrections.

henrebotha:
You can search [Cambridge](https://dictionary.cambridge.org/dictionary/english/sake?q=for+goodness%27%2C+God%27s%2C+Pete%27s%2C+heaven%27s%2C+etc.+sake), Merriam-Webster, etc. None of them have a reference for "for sth sake", it's always "for sth's sake". See also [this question](https://english.stackexchange.com/questions/101554/for-heavens-sake-or-for-heaven-sakes) on English Stack Overflow. Just like "for the graduation of my son" becomes "for my son's graduation", "for the sake of clarity" becomes "for clarity's sake". The possessive doesn't go away.

The only argument you could really make for "for sth sake" is distortion from colloquial use; but there's little reason to use such colloquial language in a technical book.

getify:
> there's little reason to use such colloquial language in a technical book.

In general, I agree, but I have indeed used a fair bit of colloquial/relaxed/informal language in these books, and many people credit that "voice" as part of what they appreciate about my books as opposed to other ones.

Since the book was already published several months ago and has already sold a few thousand copies (both print and ebook), I'm reluctant to make changes unless there's clearly a strong reason to. I couldn't argue about a misspelling or a technical inaccuracy -- those should be fixed, and I do! -- but debatable grammar choices are somewhat less clear cut.

I'll consider this change, but only if/when I find other more substantive changes to make and can roll them in at that time.

PR #1668, opened 2020-05-14
ch5 correction

uinstinct:
I think _not marked as read-only__ means **writable:true**.
Please tell me if I am incorrect!

Topic  - Setting and Shadowing Properties
Chapter 5 
Book 3 - this & Objects


getify:
This has been asked many times before. Please search existing issues. Also, as the template for PRs and the contributing guidelines assert, I'm not accepting changes to previous edition(s) of books, only the current edition (2nd-ed at present).

PR #1674, opened 2020-07-06
Misplaced punctuation

Dorni2:
I already searched for this issue 👍


getify:
I know this looks weird (does to me, too), but if you check through the book you should find that it's consistently that way -- that punctuation is inside the final quote at the end of a sentence. This is an AP style that my copy-editor applies to my work. Thanks though.

PR #1689, opened 2020-08-24
eek -> eke

philip-healy:
This is a typo fix. I have read the contributing guidelines that advise against typo fixes (specifically: "these will almost certainly be caught during the editing process"). However, I feel justified in submitting a PR for this one because it made it through the editing process to my printed copy of the second edition! Great book BTW. If I come across any other typos I will add them as additional commits to this PR.

> Typos?
>
>

----

2nd

**Book Title:**
Scope & Closures

**Chapter:**
7

**Section Title:**
Per Variable or Per Scope?

**Topic:**
Typo fix!


getify:
Good catch, thanks.

PR #1698, opened 2020-09-20
Update Ch8. Custom Change Events section. Fix calc

aaronplanell:
The original code shows:
```
obj.a;			// 12
obj.b;			// 30
obj.c;			// 3
```

But I made the calculation and the result must be:
```
obj.a;			// 6
obj.b;			// 33
obj.c;			// 3
```

Because we call to `changeObj( 3, 11 );` and:
```
	obj.a = a * 2;
	obj.b = b * 3;
```

The final value of the property `c` is OK because `a + b + old value of c` is `6 + 33 + 3` that is `42`.

Thanks!

> Typos?
>
>

----

**Book Title:**

**Chapter:**

**Section Title:**

**Topic:**

### DIFF
--- a/es6 & beyond/ch8.md
+++ b/es6 & beyond/ch8.md
@@ -241,8 +241,8 @@ Object.observe(
 
 changeObj( 3, 11 );
 
-obj.a;			// 12
-obj.b;			// 30
+obj.a;			// 6
+obj.b;			// 33
 obj.c;			// 3
 ```
 


getify:
This is an old book (not being updated after 5 years) and the API in question also doesn't exist anymore.

aaronplanell:
OK @getify . You're right, it's an old book and `Object.observe()` is obsolete right now, but the problem I commented is in regular JavaScript. Anyway, I just finished all your books. Thanks a lot for your help to the community :+1:

PR #1731, opened 2021-05-01
Ch1.md - Hoisting : Program doesn't produce error as mentioned in book and in fact hoists and return output.

purnimanagpal23:
The piece of code doesn't show reference errors anymore. In fact, it lets the variable hoist in block scope ending up showing values in the output.

**Book Title: Scope & Closures**

**Chapter: Chapter1: What's the scope**

**Section Title: Required Two Phases**

**Topic:Hoisting**

### DIFF
--- a/scope-closures/ch1.md
+++ b/scope-closures/ch1.md
@@ -133,22 +133,23 @@ Finally, consider:
 function saySomething() {
     var greeting = "Hello";
     {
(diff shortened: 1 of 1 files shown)


getify:
Your assertions and changes are not correct.

Screenshot 2021-05-01 160907

PR #1738, opened 2021-05-27
Minor changes to the README

ghost:
__Making small changes in the README, where I correct small problems of indentation, and I highlight the versions of the books to be acquired__


getify:
Thank you but I don't see any need for these changes.

PR #1758, opened 2021-09-17
Remove the phrase that was repeated twice

kumartul:


getify:
That's not a mistake, it's on purpose. It's like "I did not know that which I didn't know". AKA "unknown unknowns" if you want something to google for.

kumartul:
Oh sorry I didn't notice

PR #1760, opened 2021-09-22
get started: fixing some typos

SeyyedKhandon:


getify:
Thanks for the suggestions. This book was already published (after it received professional copy-editing) almost 2 years ago. At this point I don't want to make changes that amount to minor subjective grammatical nitpicks. I don't think these suggestions are substantive enough to warrant changes to a book already published for so long.

riverfreudenrich:
Thanks Kyle so.

River

SeyyedKhandon:
You're welcome, So does it mean, we should only focus on the other 4 books?

getify:
The way to think about this is: the bar to change/update already-published-two-years-ago books is MUCH higher than on books which are (not yet) being written. That doesn't mean no changes could ever be made, but it would require a more substantial issue (e.g., an outright mistake, bug in the code, etc). I also typically would expect to group several of those changes together and do a single update, rather than making several of those over a period of time. I would expect to update the already-published books at most once per year. But there very well might be updates that come up -- just not minor typos or grammar tweaks.

On the flip side, once I start active work on a rough draft of the book, that's also not the best time to start doing copy-editing updates, as most of that text is still very rough and subject to change, move, etc. So, the honest answer is, the best time for that kind of proof-reading effort is, during the editing/polishing phase, when a draft is being prepared for publication. I also employ professional editors to do that, but certainly am open to contributions from others concurrently.

Hopefully that helps provide some insight. :)

say2joe:
While his corrections weren't substantive enough to warrant copy changes at this time (I, IMHO, also agree that these types of changes are out-of-scope for already published works), he makes some good edits -- which are corrections. You note that you hire ("employ") professional editors -- you may want to revisit which ones you use. These grammatical errors, while small in individual increments, add up quickly to an editorial job needing an additional layer of editorial review. Just trying to politely present you with a friendly suggestion that you may want to hire or use new people for your next publication. I certainly would not use them again if they didn't catch these rather simple and obvious mistakes needing an edit. 

PS. I've been a JavaScript developer since 1995 when it debuted. Is there any possibility of contributing to any future works on the subject that you may be considering? I've also TA'd with 2U / Trilogy Education and always find errors or flaws when reading through and presenting published literature on JavaScript. I might be useful ;) But, my main intent in asking / volunteering is that I love the language and want to simply give back by helping others code correctly while using JavaScript.

getify:
I am quite happy with the quality of the editorial work that I've paid for. The editor I used (freelance) is the same editor who was employed at my original book publisher. If they're good enough for a book publisher, they're certainly good enough for me. No editor is perfect, so that's not the standard I use.

Side note: all the editing for my books is done publicly here on github, so if there's any question as to whether my editors did a good job (they did), you can actually look back through the commit history to audit that. It's also entirely possible that in porting/cherry-picking one or a handful of their edits, I missed one here or there (or just didn't agree). In fact, had my editor suggested these 2 years ago, I probably would have disagreed with many of them, at that time! But on the whole, I think their work, and mine, stands on its own in terms of quality.

I put my name on it because I'm proud of how it turned out. That doesn't mean I think it's perfect. But I take issue with the insinuation that the text is flawed or lesser as a result of the changes suggested here not being adopted. 

----

I don't think that the majority of these changes are even objectively "more correct"; IMO, most of these are judgement calls or preferential word choices.

For example: 

> who only uses ~their~ his hammer

Why should I gender that pronoun, especially in male gender? We deliberately tried to avoid a lot of gendered wording to be more inclusive. I don't agree that I should go from "their" back to "his". I *could*, but I don't think this is definitively better.

Another:

> produces ~a~ the desired outcome

Why move from "a" to "the"? The context of that sentence doesn't insist on a specific article.

And another:

> read it completely ~through~ from start to finish

Again, this is a stylistic wording choice, not an objective grammatical error.

----

Are there places where a comma should have been there to be a little more clear? Sure, I guess so. But I don't see that the majority of these nits are adding substantial improvement to my text.

I don't want to go through every single change in the PR and argue whether each is correct or not. But the fact that there are many judgements being made that I don't necessarily agree with, weakens the importance of accepting these sorts of changes as a whole set.

PR #1773, opened 2021-12-28
fix: more complete statement for Map

DBadalyan:
2-nd edition
**Book Title:**
Get Started
**Chapter:**
Chapter 3: Digging to the Roots of JS
**Section Title:**
Iteration -> Iterables
**Topic:**
Fixed statement about `Map` keys.

### DIFF
--- a/get-started/ch3.md
+++ b/get-started/ch3.md
@@ -115,7 +115,7 @@ chars;
 //   "w", "o", "r", "l", "d", "!" ]
 ```
 
-A `Map` data structure uses objects as keys, associating a value (of any type) with that object. Maps have a different default iteration than seen here, in that the iteration is not just over the map's values but instead its *entries*. An *entry* is a tuple (2-element array) including both a key and a value.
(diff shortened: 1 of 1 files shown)


getify:
The assertion in the text comes from:

1. Using a Map with primitive keys is generally wasteful, as it's probably less efficient than just using a normal object. The conventional wisdom is, use a plain object unless you need object/function keys, in which case you use a Map. I didn't explain that nuance in the text, but I agree with that opinion and based my statement on it.
2. The `WeakMap(..)` counterpart to `Map(..)` does in fact require only objects/functions as keys, so it's a frustrating gotcha if you've been in the habit of storing lots of non-object keys in Map and then you try to do so with a WeakMap and it fails.

DBadalyan:
I totally agree with your arguments, and they are indeed very important. I just disagree with an idea to have that statement in that way so it can be misused somehow via referencing.

I understand that this is not a reference book and despite that, it is even important to have very clear and complete statements, considering book's popularity among beginners.

Might be somehow it can be introduced by other way which already will emphasize notions you listed above.

Thanks for replay.

PS: not a big deal actually, just trying to be idealistic.

PR #1776, opened 2022-01-26
view shallow copy output

Jeroendevr:
Just a addition so the reader knows what the output of this shallow copy is using the spread operator. As I myself was wondering if there is a difference between var arrCopy = [ ...arr ];
or arrCopy = arr;

2
**Book Title:**
get-started
**Chapter:**
3
**Section Title:**
Iterables
**Topic:**


getify:
Not sure if I want to add this or not, but I'll keep the suggestion around to consider for future revisions.

PR #1794, opened 2022-05-20
Update ch4.md

iwestlin:
`val = yield *foo( val - 1 );` should be `return yield *foo( val - 1 );`

> Typos?
>
>

----

**Book Title:**

**Chapter:**

**Section Title:**

**Topic:**

### DIFF
--- a/async & performance/ch4.md
+++ b/async & performance/ch4.md
@@ -1491,7 +1491,7 @@ Of course, `yield`-delegation can keep following as many delegation steps as you
 function *foo(val) {
 	if (val > 1) {
 		// generator recursion
-		val = yield *foo( val - 1 );
+		return yield *foo( val - 1 );
 	}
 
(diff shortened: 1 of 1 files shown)


iwestlin:
> I am not accepting any contributions for 1st edition books.

I'm aware of this, but what about errata? Should I create a issue instead?

getify:
The proposed change is incorrect.

iwestlin:
@getify Are you sure about that? Have you run the code? Because I did...

截屏2022-05-20 下午2 23 32

截屏2022-05-20 下午2 25 09

Please kindly correct me if I was wrong or missing something...😅

iwestlin:
Oh I think I get it now, the code actually want the result of `request(...)` and pass it to another request's `?v=${val}`

I thought it just want to reduce 3 to 1 and then do the request...

Sorry for the misunderstanding, I'll close the PR right now...😢

PR #1796, opened 2022-07-16
changing repititive wording

deekeh:
"defining a definition" seems odd. It'd sound better if one of the wordings is changed.


getify:
Thanks for pointing it out. Removed some words that shouldn't have been there.

PR #1805, opened 2022-09-06
Update ch1.md

py-cs:
Fix for a minor error in destructuring example.
isDev constant should be myObj.isDeveloper, not myObj.favoriteNumber


getify:
Thank you.

PR #1806, opened 2022-09-07
Added bigint example in value type determination

### LINKED ISSUE #1790: Including bigint example in value type determination
I already searched for this issue.

**Edition:** 2nd

**Book Title:** Get started

**Chapter:** 2

**Section Title:** Value Type Determination

**Question:** As you have mentioned about `bigint`s in the chapter, shouldn't you include an example for `bigint`s, indicating that `typeof` returns `"bigint"` (and not `"number"`) for `bigint`s? that is,
```js
typeof 42;                  // "number"
typeof 42n;                 // "bigint" <-----------------
typeof "abc";               // "string"
typeof true;                // "boolean"
typeof undefined;           // "undefined"
typeof null;                // "object" -- oops, bug!
typeof { "a": 1 };          // "object"
typeof [1,2,3];             // "object"
typeof function hello(){};  // "function"
```

ajinkyac03:
I already searched for this issue.

Edition: 2nd

Book Title: Get started

Chapter: 2

Section Title: Value Type Determination

Resolves https://github.com/getify/You-Dont-Know-JS/issues/1790

### DIFF
--- a/get-started/ch2.md
+++ b/get-started/ch2.md
@@ -165,6 +165,7 @@ For distinguishing values, the `typeof` operator tells you its built-in type, if
 
 ```js
 typeof 42;                  // "number"
+typeof 42n;                 // "bigint"
 typeof "abc";               // "string"
 typeof true;                // "boolean"
 typeof undefined;           // "undefined"


getify:
Appreciate the contribution, but since this book was published 2.5 years ago (with many thousands of copies sold since), I don't see this change as important enough to warrant a deviation of the github content away from the printed book content.

PR #1824, opened 2023-03-23
Update ch3.md

eduardo-plentific:
Fixed the purpose of the spread operator in one dimension arrays (deep copies).

**Edition:** (pull requests not accepted for previous editions) - 2nd

**Book Title:** Get Started 2nd Edition

**Chapter:** - Chapter 3: Digging to the roots of JS

**Section Title:** - Consuming Iterators

**Topic:** spread operator

### DIFF
--- a/get-started/ch3.md
+++ b/get-started/ch3.md
@@ -98,7 +98,7 @@ for (let val of arr) {
 // Array value: 30
 ```
 
-Since arrays are iterables, we can shallow-copy an array using iterator consumption via the `...` spread operator:
+Since arrays are iterables, we can deep-copy a one dimension array using iterator consumption via the `...` spread operator:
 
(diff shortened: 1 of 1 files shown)


getify:
no

eduardo-plentific:
> no

So, in your example: `var arrCopy = [ ...arr ]`; does `arrCopy` become a reference of the same object in memory as `arr`?

getify:
no, it's a shallow-copy, meaning it creates a new array reference, but all its contents are the same (values or references). a deep copy means that all the values get recursively deep copied, but that's not what happens.

a "deep copy of one dimension" is just "shallow copy".

eduardo-plentific:
I see your point. Thanks!

PR #1886, opened 2024-10-16
Optimize toggle function with efficient index-based iteration.

Durgeshhirani:

**Edition:** (pull requests not accepted for previous editions) 2nd

**Book Title:** scope-closures

**Chapter:** apB

**Section Title:**  Suggested: Closure (PART 2)

**Topic:** Simpler toggle function to improve performance with index-based iteration

This PR refactors the `toggle` function to improve performance by switching from `shift()` and `push()` operations to an index-based iteration.

#### Performance Improvements:
1. **O(1) Access**: Direct index lookup (`vals[i]`) instead of `shift()` (O(n)), making it faster and more efficient.
2. **No Array Mutation**: Avoids modifying the array, reducing memory usage and ensuring more predictable performance.
3. **Efficient Iteration**: Uses a simple iterator that cycles through values, resetting when needed.

#### Benefits for Beginners:
- **Easier to Understand**: The new implementation uses basic concepts like index tracking, which is simpler and more intuitive for those learning to code.
- **Improved Performance**: The refactor helps beginners learn best practices by showcasing how to write more efficient, clean code.

### DIFF
--- a/scope-closures/apB.md
+++ b/scope-closures/apB.md
@@ -469,18 +469,15 @@ The *Closure Exercise (PART 2)* `toggle(..)` can be solved like this:
 
 ```js
 function toggle(...vals) {
-    var unset = {};
-    var cur = unset;
+    let itr = 0; // Iterator to track the current position in the values list
 
-    return function next(){
-        // save previous value back at
-        // the end of the list
-        if (cur != unset) {
-            vals.push(cur);
-        }
-        cur = vals.shift();
-        return cur;
-    };
+    return function() {
+        // If iterator exceeds or reaches the end of the list, reset to 0
+        if (itr >= vals.length) itr = 0;
+
+        // Return the current value and increment the iterator for next call
+        return vals[itr++];
+    }
 }
 
 var hello = toggle("hello");


getify:
Please see [the contributions guidelines regarding editions/books](https://github.com/getify/You-Dont-Know-JS/blob/2nd-ed/CONTRIBUTING.md#editions). This book was published (ebook/print) almost 5 years ago, with many thousands of copies sold out in the wild. It's not subject to any further changes.

----

As to the substance of your contribution, I agree it's a *valid solution*, so good job on that. FWIW, you could have used `%` to wrap around the index to remove the `if` statement.

I don't particularly agree your approach is "easier to understand", particularly to "beginners". I'd rate them approximately equal, just different.

And as to the performance difference, I suggest it probably doesn't matter at all because most of the differences are things the JS compiler will optimize, and any remaining differences will be so small that it's not relevant unless in the most extreme cases. Since performance was not at all the *domain* of discussion, I don't think it would *help* readers to delve into those sorts of optimizations, explaining or justifying one approach vs another.

Think about it this way: had I posted something like your solution, I think it's just as likely someone might have filed a PR for the opposite (the one with the "rotating queue" approach as I currently have published).

In any case, the goal of the book was to encourage you to think through closure and understand where the state needs to be held and closed-over, and you clearly got that part out, so... mission accomplished. :)

